PR #4, opened 2024-11-04
Create CODE_OF_CONDUCT.md

techcobweb:
# Why ?
We need a code of conduct spelt out at our top level.

This one looked reasonable, and I can't see any reason to diverge from it.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,128 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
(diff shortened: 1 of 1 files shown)


jt-nti (commented):
There is a [template open mainframe code of conduct](https://github.com/openmainframeproject/project-template/blob/main/CODE_OF_CONDUCT.md) which unfortunately leads to a bit of a dead end _but_ it does include a contact email which seems useful... 

> Please contact [conduct@openmainframeproject.org](mailto:conduct@openmainframeproject.org) to report any violations or concerns.

If the address is not still correct, it would be useful to include some Galasa-specific address to report concerns to.

techcobweb:
The TSC project already has one. Need to sue that, or get it to refer to this one.

techcobweb:
stale

PR #7, opened 2024-11-06
Move auth-related functions into dev.galasa.framework.auth.spi bundle

eamansour:
## Why?
Related to https://github.com/galasa-dev/projectmanagement/issues/1975

## Changes
- Added new `dev.galasa.framework.auth.spi` bundle to house common auth-related functionality for the auth and users servlets to use


aashir21 (approved):
👍🏻

modules/framework/galasa-parent/dev.galasa.framework.api.authentication/src/main/java/dev/galasa/framework/api/authentication/internal/routes/AuthTokensDetailsRoute.java
```diff
@@ -58,7 +50,7 @@ public HttpServletResponse handleDeleteRequest(String pathInfo, QueryParameters
     private String getTokenIdFromUrl(String pathInfo) throws InternalServletException {
         try {
             // The URL path is '/auth/tokens/{tokenid}' so we'll grab the {tokenid} part of the path
-            Matcher matcher = Pattern.compile(PATH_PATTERN).matcher(pathInfo);
+            Matcher matcher = this.getPath().matcher(pathInfo);
```
techcobweb: It's a regex which is returned, so call it `getPathRegex()` ?

techcobweb (approved):

techcobweb:
Having a static method for the factory - that will cause clashes between test cases who try to set up a different thing being returned ?

You could have the factory implement an interface, and call static getInstance() to get the factory interface. Test code can then create a mock factory, and pass the interface to code ?

But really, the OSGi service dance would provide this abstraction from both statics and from new-ing up concrete classes.

PR #8, opened 2024-11-06
mcobbett add dependabot rule

techcobweb:
- **Added gradle libs.versions.toml file to guide gradle about versions of libraries we depend upon**
- **tidy up a build.locally script slightly.**
- **creating a libs.versions.toml file to control versions of gradle packages using a version catalog**
- **add the bundle for log4j libraries**


eamansour (approved):

techcobweb:
We went with a gradle platform rather than this toml file approach.

PR #21, opened 2024-11-15
Move extensions.mocks into extensions.common as a test fixture to remove cyclic dependency

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2055

Removes the cyclic dependency between extensions.common and extensions.mocks by moving the mocks into a test fixture within extensions.common and updating all consumers to use this test fixture


modules/extensions/galasa-extensions-parent/dev.galasa.cps.rest/build.gradle
```diff
@@ -17,9 +17,7 @@ dependencies {
     // https://mvnrepository.com/artifact/org.apache.httpcomponents/httpclient
     implementation 'org.apache.httpcomponents:httpclient:4.5.14'
 
-    testImplementation 'junit:junit:4.13.1'
-    testImplementation(project(':dev.galasa.extensions.mocks'))
-    testImplementation 'org.assertj:assertj-core:3.16.1'
```
jadecarino, line 22: Are junit and assertj-core not needed anymore?

eamansour: They're included in the "galasa.extensions" plugin - these were duplicated in this build.gradle

jadecarino (commented):
Hi Eamonn, looks great thanks just a couple of questions -

Are junit and assertj-core not required anymore in cps.rest where they have been removed?

And also does the set-version.sh script now need to update the dev.galasa.framework version in its new location (dev.galasa.extensions.common)?

modules/extensions/galasa-extensions-parent/dev.galasa.cps.rest/build.gradle
jadecarino: Great thanks

eamansour:
> Hi Eamonn, looks great thanks just a couple of questions -
> 
> Are junit and assertj-core not required anymore in cps.rest where they have been removed?
> 
> And also does the set-version.sh script now need to update the dev.galasa.framework version in its new location (dev.galasa.extensions.common)?

Thanks for the review!

See https://github.com/galasa-dev/galasa/pull/21#discussion_r1843794158 for my comment on the junit and assertj-core changes.

The set-version script in extensions already checks for `dev.galasa.framework` references as well as the build.gradle `version = xxx` lines, and updates both (see https://github.com/galasa-dev/galasa/blob/main/modules/extensions/set-version.sh#L106-L109). Just tested it out and it seems to work as we'd expect :)

Since we'll be adopting the platform and using the same Galasa version across bundles, we should really be grabbing the framework version from a global property, something like:

```groovy
ext {
    GALASA_VERSION = "0.38.0"
}
```

Maybe that could be added as part of the platform story?

techcobweb:
@eamansour 
> Since we'll be adopting the platform and using the same Galasa version across bundles, we should really be grabbing the framework version from a global property, something like...etc.

I have been working on this. [Infrastructure: Tool to update the release version dynamically wherever it is referenced. #2053](https://github.com/galasa-dev/projectmanagement/issues/2053)

...and am doing something similar to what you suggest, though more like this:
```
allprojects {
    group = 'dev.galasa'
    version = "0.39.0"
}
```
... so that anything which currently uses the `version` variable can do so. Is the `ext` method better then ?

techcobweb (approved):

eamansour:
> I have been working on this. https://github.com/galasa-dev/projectmanagement/issues/2053

That's great, thanks for pointing that out.

>  Is the `ext` method better then ?

Using `ext` would mean having to point the `group` and `version` properties to values like `ext.GALASA_VERSION` and `ext.GALASA_GROUP`, which isn't as seamless as just setting a global `group` and `version` like the `allprojects` method does. I'd be more than happy with using `allprojects` if it works!

PR #27, opened 2024-11-18
wrapping module to use the gradle platform to get version numbers

techcobweb:

# Why ?

We need to use the centralised platform to get our version numbers in the wrapping module.

This is the initial work at getting some of the version numbers into the platform, so wrapping can use them.

- Update the OBR release.yaml
- Update the OBR set-version.yaml so that the version of the wrapping bundles is always uplifted

- Create a set-version script for the platform module
- Create an overall set-version script which calls the smaller module ones.
- Change the wrapping auth0 bundle to have the dev.galasa.wrapping prefix, and set it's version to 0.38.0 so it should always bump up from now on also.
- Change all references to the auth0 bundle to use the new name.

- An attempt at the GH Actions flow to pull down the platform built artifacts so the wrapping project has access to the BOM.


modules/framework/run-locally.sh
```diff
@@ -123,7 +123,7 @@ OBR_VERSION="0.31.0"
 # version = "0.37.0"
 
 framework_version=$(cat $BASEDIR/galasa-parent/dev.galasa.framework/build.gradle | grep "version =" | cut -f2 -d'"' | xargs)
-OBR_VERSION="$framework_version"
+OBR_VERSION="0.31.0"
```
jadecarino: Should this be 0.38.0?

modules/wrapping/pom.xml
```diff
@@ -12,6 +12,15 @@
 	<description>Wraps non-OSGi JARs into OSGi Bundles for use in the Galasa Ecosystem</description>
 	<url>https://galasa.dev</url>
 
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+		<java.version>1.8</java.version>
+		<maven.compiler.source>1.8</maven.compiler.source>
+		<maven.compiler.target>1.8</maven.compiler.target>
```
jadecarino: Should these be a more up to date version?

jadecarino (commented):
I have just a couple questions on versions used here but so far looks good

modules/framework/run-locally.sh
techcobweb: No it should probably be a variable controlled by the set-version tool.

modules/wrapping/pom.xml
techcobweb: Perhaps we should target Java 11 bytecode now yes. Probably was wider than this one place in the code tho. Will see...

techcobweb: See https://stackoverflow.com/questions/38882080/specifying-java-version-in-maven-differences-between-properties-and-compiler-p
for background info. Seems there is a better way of specifying this now.

techcobweb: OK. I applied that to the wrapping and maven plugin modules.
Checked that bytecode was being compiled to java 11 format rather than Java 1.8
(major version: 0, minor version: 55). https://stackoverflow.com/questions/9170832/list-of-java-class-file-format-major-version-numbers

techcobweb: Fixed the set-version.sh script, and the run-locally.sh script. API server works with the web UI on my local machine now.

jadecarino (approved):
Looks good to me - approved

PR #31, opened 2024-11-18
Iss 2052 - Use the Java Gradle Platform in the Managers module

jadecarino:
## Why?

Centralise versions of all dependencies used in the Galasa Managers. For https://github.com/galasa-dev/projectmanagement/issues/2052

- Update build.gradle files to pull version of dependencies from the platform.
- Update bnd.bnd files to pull version of jars from the build process.
- Upgraded all manager bundles to 0.38.0 (other than managers due to be removed from code base).
- OBR release.yaml updated with new commons-codec and assertj-core versions, and reordered to be in alphabetical order to make easier side-by-side comparisons with the platform in future.
- Tested running the API Server locally and ran all available IVTs to check there are no wiring issues.
- A few dependencies have been left with their hard coded version (mockito-core, okhttp and kotlin-osgi-bundle) as their versions clash with the platform's so these will be addressed and necessary changes made at later point.


modules/managers/galasa-managers-parent/galasa-managers-cloud-parent/dev.galasa.docker.manager/build.gradle
```diff
@@ -4,14 +4,15 @@ plugins {
...
-version = '0.34.0'
+version = '0.38.0'
 
 dependencies {
     implementation project(':galasa-managers-comms-parent:dev.galasa.http.manager')
     implementation project(':galasa-managers-core-parent:dev.galasa.artifact.manager')
-    implementation 'com.google.code.gson:gson:2.10.1'
-    implementation 'commons-io:commons-io:2.16.1'
-    implementation 'org.apache.commons:commons-compress:1.26.0'
+    implementation platform('dev.galasa:dev.galasa.platform:0.38.0')
```
eamansour: Since we've already added the platform to the galasa.manager plugin (which all managers use), do we still need to add the platform into the build.gradle file for each manager? Would be great to check if the `implementation platform` lines can be removed from the build.gradle files for the managers.

The same applies throughout this PR and for IVTs (as we've added the platform into the galasa.manager.ivt plugin as well).

eamansour (approved):
One question but otherwise looks good!

modules/managers/galasa-managers-parent/galasa-managers-cloud-parent/dev.galasa.docker.manager/build.gradle
jadecarino: Great point I'll remove those lines and see if all still happy

jadecarino: Everything works with the `implementation platform` lines being removed from the individual bundles

PR #39, opened 2024-11-19
jta bundle re-named and re-versioned

techcobweb:

# Why ?
Edging towards full adoption of the gradle platform construct to control versions of things.

This PR changes the `jta` bundle in wrapping such that it gets re-named, and re-versioned to be the same version as the release.


jadecarino (approved):
Looks good to me

PR #45, opened 2024-11-20
Adding in Tolerations to the Galasa TestPodScheduler

Tom-Slattery:
The Galasa test pod scheduler can handle Node Affinities, for our usage in CICS we also need it to be able to handle Node Tolerations, so that we can make a node exclusively run our Galasa test pods.

I have added a property to the Settings.java file for Node Tolerations based on how the Node Affinity property is handled. This should allow you to specify multiple node tolerations in your Ecosystem Config as a list:

```
  galasa_node_tolerations: "node-label=Operator:Condition,node-label2=Operator2:Condition2..."
```
for example:
```
  galasa_node_tolerations: "galasa-engines=Exists:NoSchedule"
```

I have added code to the TestPodScheduler that will process the text as a list and add them as tolerations items in the podspec before creating the pod.

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
@@ -277,6 +277,16 @@ public class TestPodScheduler implements Runnable {
         return newPod;
     }
 
+
+    /*
+    * Tolerations are supplied as a string in the form:
+    * "node-label1=Operator1:Condition1,node-label2=Operator2:Condition2"
+    *
+    * For example: "galasa-engines=Exists:NoSchedule"
+    *
+    * The following method parses the String comma separated list of node
+    * tolerations and returns a list of K8s V1Tolerations.
+    */
     private List<V1Toleration> createNodeTolerations(String nodeTolerations) {
         List<V1Toleration> tolerationsList = new ArrayList<>();
 
--- a/modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
@@ -268,8 +268,10 @@ public class TestPodScheduler implements Runnable {
 
             if(tolerationsList.length > 0) {
                 for(int i = 0; i < tolerationsList.length; i++){
-                    String[] selection = nodePreferredAffinity.split("=");
+                    String[] selection = tolerationsList[i].split("=");
+
                     if (selection.length == 2) {
+
                         String[] operatorAndEffect = selection[1].split(":");
 
                         if(operatorAndEffect.length == 2) {
--- a/modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
@@ -259,46 +259,54 @@ public class TestPodScheduler implements Runnable {
                 requirement.setKey(selection[0]);
                 requirement.setOperator("In");
                 requirement.addValuesItem(selection[1]);
+
+
             }
         }
 
         String nodeTolerations = this.settings.getNodeTolerations();
-        if (!nodeTolerations.isEmpty()) {
-            String[] tolerationsList = nodeTolerations.split(",");
-
-            if(tolerationsList.length > 0) {
-                for(int i = 0; i < tolerationsList.length; i++){
-                    String[] selection = tolerationsList[i].split("=");
-
-                    if (selection.length == 2) {
-
-                        String[] operatorAndEffect = selection[1].split(":");
-
-                        if(operatorAndEffect.length == 2) {
-                            V1Toleration toleration = new V1Toleration();
-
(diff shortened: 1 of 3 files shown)


Tom-Slattery:
I have installed a new Galasa Ecosystem at the latest "main" version. I've edited the Galasa boot image to include my changes to the k8s controller JAR. I can confirm that the ecosystem consistantly schedules the test runs on the correct node with this Tolerations change, which would not otherwise be possible without the node toleration.

modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
```diff
@@ -261,6 +262,40 @@ V1Pod createTestPod(String runName, String engineName, boolean isTraceEnabled) {
             }
         }
 
+        String nodeTolerations = this.settings.getNodeTolerations();
```
techcobweb, line 267: can this pls be put into a separate method, as it's getting pretty big...

techcobweb: Also, any unit tests for that method would be good ?

techcobweb:
Also, some documentation on how an ecosystem admin is supposed to use this new feature would be needed. Where ? Not sure. Even if it's in a separate readme.md somewhere in the helm chart perhaps ?

modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
Tom-Slattery: Hi, moved the nodetoleration creation into it's own method. I've also added a unit test for the nodetolerations creation. I spotted that the nodeaffinity creation didn't have a unit test already, so I've added one for that too.

Tom-Slattery:
> Also, some documentation on how an ecosystem admin is supposed to use this new feature would be needed. Where ? Not sure. Even if it's in a separate readme.md somewhere in the helm chart perhaps ?

Here is some documentation about the expected usage of this feature for an ecosystem admin:



_As an ecosystem admin, you may need to configure where which hardware nodes your Galasa test pods are run on. As an example, you may wish to dedicate certain k8s nodes to exlusively to running Galasa tests to avoid resource conflicts with other workloads on the cluster. This ensures hardware isolation and gives a better guarantee of the performance of your Galasa Ecosystem._

_Two components are required to achieve this:_
_- A k8s node affinity allows you to designate a preffered node for your Galasa tests to be run on_
_- A k8s node taint prevents other workloads from scheduling on that node. To allow you Galasa test pods to be scheduled on the tainted node, you must define a node tolerance_

_These two settings can be configured via the Galasa Ecosystem configmap, for example:_

```
galasa_node_preferred_affinity: galasa-engines

galasa_node_tolerations: galasa-engines=Exists:NoSchedule
```

_The node affinity defines a label that can be used to identify nodes that your Galasa Ecosystem should schedule pods on._ 

_The Galasa Node Tolerations param defines a comma-separated list of toleration conditions of the format "nodelabel=Operator:Condition". In the above example, nodes with the label "galasa-engines" that have a "NoSchedule" taint should be tolerated, allowing the Galasa test run pods to be scheduled on that node._


Where do we want this to be included? Is there somewhere that documents the usage of the other parameters of the Galasa Ecosystem configmap?

techcobweb:
Can you check that into an .md file please ? Then at least we can gather it into our docs when @jt-nti has time ?

techcobweb:
Should these settings be surfaced into the values.yaml in the helm repo ? In which case we need a separate helm repo PR with them in. Which is my preferred place to put this level of detail.

techcobweb:
The docs also need to say what happens if they are missing, and that these parameters are optional.

Tom-Slattery:
> Should these settings be surfaced into the values.yaml in the helm repo ? In which case we need a separate helm repo PR with them in. Which is my preferred place to put this level of detail.

Raised a PR here to add doc and params to the values.yaml file for the ecosystem deployment:

https://github.com/galasa-dev/helm/pull/56

modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
```diff
@@ -266,6 +277,38 @@ V1Pod createTestPod(String runName, String engineName, boolean isTraceEnabled) {
...
+        List<V1Toleration> tolerationsList = new ArrayList<>();
+
+        String[] tolerationStringSplit = nodeTolerations.split(",");
+
+        if(tolerationStringSplit.length > 0) {
+            for(int i = 0; i < tolerationStringSplit.length; i++){
+                String[] selection = tolerationStringSplit[i].split("=");
+
+                if (selection.length == 2) {
+                    String[] operatorAndEffect = selection[1].split(":");
```
eamansour, line 300: A comment showing the expected format of the provided tolerations list might make this easier to read in the future

eamansour (approved):
One small comment but approved. Thanks!

modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
Tom-Slattery: Does this need to be included in the code or is this covered by the documentation in the helm chart here https://github.com/galasa-dev/helm/pull/56 ?

Tom-Slattery: Just pushed a commit with a comment describing the format expected

PR #47, opened 2024-11-20
Created a new detect secrets script

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2047


modules/managers/build-locally.sh
```diff
@@ -250,8 +250,20 @@ function update_release_yaml {
...
+        TOOL_DIR=$(pwd)
+        $TOOL_DIR/tools/detect-secrets.sh
+    fi
+}
+
 
 build_code
 update_release_yaml
 
-check_secrets
```
techcobweb, line 257: It was previously clcear that check-secrets was being called, now it's a side-effect of checking something ?
Suggest you just call the check_if_script_invoked_directly back to check_secrets_unless_supressed or something more meaningful.

modules/managers/build-locally.sh
```diff
@@ -250,8 +250,20 @@ function update_release_yaml {
     fi
 }
 
+function check_if_script_invoked_directly() {
+    # Check if the script is being called directly or from another script
+    if [[ -z "${IN_CHAIN_MODE}" ]]; then
+        info "Script invoked directly, running detect-secrets.sh script"
+
+        # Run the detect-secrets.sh in root
+        cd "${WORKSPACE_DIR}/.."
```
techcobweb: Shouldn't WORKSPACE_DIR refer to the repo root ?
Then cd into the tools dir, so "/tools" is part of the TOOL_DIR ?

modules/managers/build-locally.sh
```diff
@@ -250,8 +250,20 @@ function update_release_yaml {
...
+        TOOL_DIR=$(pwd)
+        $TOOL_DIR/tools/detect-secrets.sh
```
techcobweb: odd seeing the `tools` when you already have a TOOL_DIR ?

modules/maven/build-locally.sh
```diff
@@ -105,6 +105,18 @@ function check_secrets {
     success "secrets baseline timestamp content has been removed ok"
 }
 
+function check_if_script_invoked_directly() {
```
techcobweb: As this is used everywhere, why not just call the function in the main tools folder from everywhere ?

modules/managers/build-locally.sh
```diff
@@ -250,8 +250,20 @@ function update_release_yaml {
     fi
 }
 
+function check_if_script_invoked_directly() {
+    # Check if the script is being called directly or from another script
+    if [[ -z "${IN_CHAIN_MODE}" ]]; then
```
techcobweb: I prefer the script to have a new flag. `--detect-secrets true` or `--detect-secrets false` with the default to be true, but when the uber-build-tool calls it, it can use false ? Rather than this hidden IN_CHAIN_MODE env variable...

.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
+  - repo: https://github.com/ibm/detect-secrets
+    rev: 0.13.1+ibm.62.dss
+    hooks:
+      - id: detect-secrets # pragma: whitelist secret
+        args: [--baseline, .secrets.baseline, --use-all-plugins, --fail-on-unaudited]
```
techcobweb, line 8: How does someone hook this in ?

tools/build-locally.sh
```diff
@@ -197,7 +197,7 @@ function build_module() {
     if [[ "$module" == "gradle" ]]; then
         h2 "Building $module"
         cd ${PROJECT_DIR}/modules/$module
-        ${PROJECT_DIR}/modules/$module/build-locally.sh
+        IN_CHAIN_MODE=true ${PROJECT_DIR}/modules/$module/build-locally.sh
```
techcobweb: prefer a parameter being supported by each of the smaller .sh files. rather than pass an environment variable.

That way, it gets documented in the usage() text of the smaller build-locally.sh tool.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
```
techcobweb, line 3: Some comment in here about what this file is trying to do and how it does it would be good.

techcobweb (approved):
Aproved

techcobweb:
And you have clashes from main which need merging.

PR #58, opened 2024-11-22
Bump commons-logging to 1.3.4 and override commons-logging for the httpclient-osgi bundle

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2028

## Changes
- Bumped commons-logging to 1.3.4
- Replaced httpclient-osgi bundle with a custom wrapped bundle in order to override the version of commons-logging in that bundle since it was constrained to use commons-logging between 1.1.0 and 1.3.0 (this previously caused wiring errors when bumping commons-logging to 1.3.4)


jadecarino (approved):
Good stuff and well done for tracking down the reason why commons-logging was originally removed from the OBR! Comment will be helpful going forward. Approved!

PR #61, opened 2024-11-25
Extensions modules all have the same version controlled from the top level.

techcobweb:

# Why ?
As See story https://github.com/galasa-dev/projectmanagement/issues/2053

In this PR we change the top-level version and it gets pushed to all the sub-projects in the extensions module.


modules/obr/dependency-download/build.gradle
```diff
@@ -46,6 +47,14 @@ task getFrameworkDependency {
     }
 }
 
+// Gets the framework dependency and renames it to remove the version number.
```
jadecarino: I think this is meant to say extensions here

jadecarino (approved):
Approved - although there is one comment you might want to update

modules/obr/dependency-download/build.gradle
techcobweb: fixed

PR #69, opened 2024-11-25
Added detect-secrets to workflow

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2047

### DIFF
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -65,6 +65,14 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: |
           ./tools/get-last-successful-workflow-run-for-artifacts.sh --repo ${{ github.repository }}
+          
+      - name: Turn script into an executable
+        working-directory: ./tools
+        run: chmod +x detect-secrets.sh
+    
+      - name: Run the detect secrets script
+        working-directory: ./tools
+        run: detect-secrets.sh
 
   pr-build-platform:
     name: Build the 'platform' module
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-requests.yaml
```diff
@@ -65,6 +65,14 @@ jobs:
...
         run: |
           ./tools/get-last-successful-workflow-run-for-artifacts.sh --repo ${{ github.repository }}
+          
+      - name: Turn script into an executable
+        working-directory: ./tools
+        run: chmod +x detect-secrets.sh
+    
+      - name: Run the detect secrets script
+        working-directory: ./tools
+        run: ./detect-secrets.sh
```
jadecarino: Please can you create a new job called `detect-secrets` at the top of the workflow and put these two steps in there. The job `end-pull-request-build` will need `detect-secrets` added to its "needs" section - so the pull request cannot end successfully unless that job was successful.

jadecarino (changes requested):
Hi Ash, so GitHub Actions workflows are structured in a hierarchy of workflow > jobs > steps. In this workflow there are the jobs `get-changed-modules`, `find-artifacts`, and then several `pr-build-*` which are jobs that call other workflows (known as reusable workflows). The two steps you have added have been put into the `find-artifacts` job which would work, but they should be put in their own job so that 1 each job does a specific thing and 2 jobs can run in parallel to eachother (steps cannot) so the detect-secrets can run while other things are happening.

Please move the steps into their own job as described in the comment.

jadecarino (approved):
Good stuff :)

PR #76, opened 2024-11-27
Fix: Don't use platform to generate the galasa-bom

jadecarino:
## Why?

The galasa-bom is a component used by Galasa user's to get versions of dependencies for their test projects (generated by `galasactl project create`). Using the platform to generate the galasa-bom subsequently broke the `project create` command so I have reverted the changes to the galasa-bom where the platform was involved (found in this [PR](https://github.com/galasa-dev/galasa/pull/70/files#diff-326cccbabfee54a9f05abac16cfef22e42b4eb13a834ae0e717a29d2a2a3cc54))
The OBR's release.yaml now requires all dependencies with `bom: true` to list their version explicitly again. So I have added a comment in the platform to highlight the second location for uplifts in future.

### DIFF
--- a/modules/obr/galasa-bom/pom.template
+++ b/modules/obr/galasa-bom/pom.template
@@ -42,26 +42,16 @@
     </issueManagement>
 
     <dependencyManagement>
-		<dependencies>
-			<dependency>
-				<groupId>dev.galasa</groupId>
-				<artifactId>dev.galasa.platform</artifactId>
-				<version>0.38.0</version>
-				<type>pom</type>
-				<scope>import</scope>
-			</dependency>
-		</dependencies>
-	</dependencyManagement>
-
-    <dependencies>
+        <dependencies>
 {{range .Artifacts}}
         <dependency>
             <groupId>{{.GroupId}}</groupId>
-            <artifactId>{{.ArtifactId}}</artifactId>{{if .Version}}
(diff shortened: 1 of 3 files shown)


modules/obr/galasa-bom/pom.template
```diff
@@ -42,26 +42,16 @@
...
-
-    <dependencies>
+        <dependencies>
 {{range .Artifacts}}
         <dependency>
             <groupId>{{.GroupId}}</groupId>
-            <artifactId>{{.ArtifactId}}</artifactId>{{if .Version}}
-            <version>{{.Version}}</version>{{end}}
+            <artifactId>{{.ArtifactId}}</artifactId>
+            <version>{{.Version}}</version>
```
techcobweb, line 50: We are in danger of having `<version></version>` in the bom now, if someone misses the version of something and says it should be in the bom ?

Wondering if we should error in some way instead if version isn't set ?

techcobweb (approved):

modules/obr/galasa-bom/pom.template
jadecarino: I can update the galasabld code which generates the bom to check each artifact for if bom is true, then version must not be blank, if it is then error. I'll deliver this PR to unblock others then will look at the galasabld code after.

PR #79, opened 2024-11-27
mcobbett framework version set from top

techcobweb:
- **set version of all framework bundles at the top**
- **maven plugin versions are bumped every release.**


modules/framework/galasa-parent/galasa-demo-archetype/pom.xml
```diff
@@ -6,11 +6,11 @@
 	<parent>
 		<groupId>dev.galasa</groupId>
 		<artifactId>galasa-parent</artifactId>
-		<version>0.15.0</version>
+		<version>0.38.0</version>
 	</parent>
 	
 	<artifactId>galasa-demo-archetype</artifactId>
-	<version>0.15.0</version>
+	<version>0.38.0</version>
```
jadecarino, line 13: I don't think this is used anywhere so it might be worth confirming then deleting so we don't release anything unnecessary in the 0.38.0 release - but that can be done after this PR is merged

jadecarino (approved):

modules/framework/galasa-parent/galasa-demo-archetype/pom.xml
techcobweb: Raised separate story to look at this.

PR #83, opened 2024-11-28
Iss2052 - Add platform to more wrappers

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2052

All dependencies used in the wrappers with a consistent version across the codebase (either they're only used once or all other uses of the dependency share the same version) have been put into the platform and their hard coded versions removed.

Hard coded version of the jsr305 jar mentioned in bnd.bnd file of the selenium manager removed as it can now get its version from the platform.

Dependencies with conflicting versions across the codebase added to follow up story.

Testing done:
1. galasa repo built
2. cli repo built and generated test project ran
3. CoreManagerIVT ran locally
4. CoreLocalJava11Ubuntu ran locally


techcobweb (approved):
nice

PR #84, opened 2024-11-28
Iss2052 - Add platform to extensions

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2052

All dependencies used in the extensions with a consistent version across the codebase (either they're only used once or all other uses of the dependency share the same version) have been put into the platform and their hard coded versions removed.

bnd.bnd file for cps.etcd updated with wildcards to find the version on the classpath.

Dependencies with conflicting versions across the codebase added to follow up story.

Testing done:
- [x] galasa repo built
- [x] cli repo built and generated test project ran
- [x] CoreManagerIVT ran locally
- [x] CoreLocalJava11Ubuntu ran locally


modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/build.gradle
```diff
@@ -6,27 +6,27 @@ plugins {
...
     // }
     // We don't want to use their gson implementation.
-    implementation ('com.google.code.gson:gson:2.10.1')
+    implementation ('com.google.code.gson:gson')
     
     // Not required for compile,  but required to force the download of the jars to embed by bnd
     implementation('com.fasterxml.jackson.core:jackson-core:2.16.1')
     implementation('com.google.api.grpc:proto-google-common-protos:2.41.0')
-    implementation('com.google.code.gson:gson:2.10.1')
+    implementation('com.google.code.gson:gson')
```
eamansour: Looks like gson was previously added in here twice - would be good to remove the duplicate one

eamansour (approved):
One minor comment but approved!

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/build.gradle
jadecarino: Removed duplicate dependency

PR #85, opened 2024-11-28
Added fix for casing while storing runs

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2056

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/FrameworkRuns.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/FrameworkRuns.java
@@ -242,7 +242,28 @@ public class FrameworkRuns implements IFrameworkRuns {
         } else {
             otherRunProperties.put(runPropertyPrefix + ".group", UUID.randomUUID().toString());
         }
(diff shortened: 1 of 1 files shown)


modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/FrameworkRuns.java
```diff
@@ -242,7 +242,28 @@ private boolean storeRun(String runName, SubmitRunRequest runRequest) throws Dyn
...
+        //If set to true then it will allow it to have mixed characters, for e.g,
+        // Galasadelivery@ibm.com. Else it would set everything to lowercase by default
+
+        boolean isMixedCaseRequestorAllowed = false;
+        try {
+            String propValue ;
+            propValue = cps.getProperty("runs.requestor", "is.mixed.case");
+            if ("true".equals(propValue)) {
+                isMixedCaseRequestorAllowed = true;
+            }
```
eamansour, line 256: `Boolean.parseBoolean()` could be used to handle inputs like `True`, `TRUE`, and `tRuE`.

Since this is a feature flag, I'm happy to keep as is.

eamansour (approved):

PR #90, opened 2024-11-28
Remove the MQ manager from being an optional resource for the framework bundle

eamansour:
## Why?
When the framework bundle is installed, the MQ manager incorrectly appears as an optional resource. This has led to several regression test failures as the MQ manager is not included in our MVP packaging. This PR fixes this erroneous behaviour so that the MQ manager is not listed as an optional resource when the framework is installed.


jadecarino (approved):
Great find :)

PR #111, opened 2024-12-06
galasabld now builds with linux arm64 architecture

wjbartol:
Running a docker container on a mac requires the galasabld tool to support the linux-arm64 architecture.


techcobweb:
We've done this now in a separate PR.

techcobweb:
Looks like this change has already been made ? Closing as a result. Please re-open if that's not the case.

PR #115, opened 2024-12-10
Iss2082 - Build image for release process with OBR and a galasabld executable

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2082

As part of the release, to release the dev.galasa artifacts to the staging repo, we call a script -> which calls a Tekton pipeline -> which runs the `galasabld maven deploy tool`. To do this, we need the Docker image that runs in the pipeline to have two things installed: 1. all of the OBR artifacts and 2. the galasabld executable. This image should be built as part of the release process. Currently it's not and so it was built manually for the 0.38.0 release.

### DIFF
--- a/modules/obr/dockerfiles/dockerfile.obrgalasabld
+++ b/modules/obr/dockerfiles/dockerfile.obrgalasabld
@@ -5,6 +5,6 @@ ARG tag
 FROM ${dockerRepository}/galasa-dev/obr-maven-artefacts:${tag}
 
 ARG platform
-COPY bin/galasabld-${platform} /bin/galasabld
+COPY galasabld-${platform} /bin/galasabld
 
 ENTRYPOINT ["/bin/galasabld"]
--- a/modules/obr/dockerfiles/dockerfile.obrgalasabld
+++ b/modules/obr/dockerfiles/dockerfile.obrgalasabld
@@ -1,9 +1,15 @@
+# During the release process, we need to be able to execute `galasabld`
+# commands from within an image that contains all the 'dev.galasa' artifacts.
+# This is not a hosted image - but it runs a pipeline which copies the
+# artifacts from the image's filesystem to a Sonatype staging repo over HTTP.
 ARG dockerRepository
 ARG tag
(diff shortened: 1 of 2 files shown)


modules/obr/dockerfiles/dockerfile.obrgalasabld
```diff
@@ -5,6 +5,6 @@ ARG tag
 FROM ${dockerRepository}/galasa-dev/obr-maven-artefacts:${tag}
```
techcobweb, line 10: Could you pls add a description of what this docker image is trying to do ?
eg: 

An image which hosts a maven site with apache, where the maven site contains
all the maven artifacts from the OBR build, plus the galasactl command-line tool (all architectures/platforms).

Or something like that ?

Also, why do we need this ?
eg:
This is hosted, (where?) so that people can download the artifacts served up.

Or something like that ? Is that correct ? or is it the build process which needs it ?

techcobweb:
> which calls a Tekton pipeline

So our release process still needs Tekton ?
Which stories do we need to complete before that isn't the case pls ?

techcobweb:
> This image should be built as part of the release process

Shouldn't it always be built ? So we know that this part of the build always builds, rather than find out when we just do the release process ?

techcobweb (approved):

modules/obr/dockerfiles/dockerfile.obrgalasabld
jadecarino: Comment added for clarity.

jadecarino:
> > which calls a Tekton pipeline
> 
> So our release process still needs Tekton ? Which stories do we need to complete before that isn't the case pls ?

Yes just the release process. It's story https://github.com/galasa-dev/projectmanagement/issues/2087

jadecarino:
> > This image should be built as part of the release process
> 
> Shouldn't it always be built ? So we know that this part of the build always builds, rather than find out when we just do the release process ?

I have updated this to always build.

PR #117, opened 2024-12-11
set the version into the build workflows so the build logic knows which version it is building

techcobweb:

# Why ?
Part of this story: https://github.com/galasa-dev/projectmanagement/issues/2053

- Version master sits in the `/build.properties` file where GALASA_VERSION is set.
- The workflows ~~all~~ which need it go get this file (using sparse checkout), and convert the value into the `galasa-version` variable
- This can then be passed down into sub-flows
- The set-version.sh script sets the version into the `/build.properties` file.

### DIFF
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -11,6 +11,21 @@ on:
 
 jobs:
 
+  set-build-properties:
+    # Use the version.properties file to source the galasa-version output variable.
+    # This variable can be referenced by other jobs in this flow using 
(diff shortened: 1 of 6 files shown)


.github/workflows/pull-requests.yaml
```diff
@@ -11,6 +11,21 @@ on:
 
 jobs:
 
+  set-build-properties:
```
jadecarino: Shall we remove this from here as the output isn't passed into pr-obr.yaml as an input so nothing is done with the galasa-version here

jadecarino (approved):
Just one comment but otherwise looks good!

PR #108, opened 2024-12-06
Use async requests to delete runs and add view to get runs by run name more efficiently

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2020

## Changes
- Deleting all associated documents of a run is now done asynchronously to speed up deletion
- Added a new run name view to get runs by run name (which is faster than using a mango query)
  - Changed the `GET /ras/runs` route to use this new view when getting runs by run name

Local testing shows an improvement in both `runs get --name xxx` and `runs delete`.

Previously:
```
❯ time galasactl runs get --name C11523
submitted-time(UTC) name   requestor              status   result test-name
2024-09-27 15:30:20 C11523 galasadelivery@ibm.com finished Passed dev.galasa.inttests.core.local.CoreLocalJava11Ubuntu

Total:1 Passed:1
galasactl runs get --name C11523  0.00s user 0.01s system 0% cpu 1.734 total

❯ time galasactl runs delete --name C11523
galasactl runs delete --name C11523  0.00s user 0.01s system 0% cpu 2.002 total
```

Now:
```
❯ time galasactl runs get --name C11522
submitted-time(UTC) name   requestor              status   result test-name
2024-09-27 15:19:50 C11522 galasadelivery@ibm.com finished Passed dev.galasa.inttests.core.local.CoreLocalJava11Ubuntu

Total:1 Passed:1
galasactl runs get --name C11522  0.00s user 0.01s system 14% cpu 0.091 total

❯ time galasactl runs delete --name C11522
galasactl runs delete --name C11522  0.00s user 0.00s system 7% cpu 0.109 total
```

PR #124, opened 2024-12-13
Only delete run document after all referenced documents are deleted

eamansour:
## Why?
Related to changes in https://github.com/galasa-dev/galasa/pull/108

With the async requests, deleting runs currently deletes the parent run document at the same time as it deletes referenced artifact and log documents. If the deletion of a document fails (e.g. couchdb or the API server goes down) and the parent run document has already been deleted, it may cause some artifact documents to be orphaned.

## Changes
- Make sure that the parent run document is only deleted **after** all referenced artifact and log documents have been deleted
- Ignore 404 Not Found errors when deleting documents since we should only error out if any server errors occur
- Change request origin not set log message from error to info since it doesn't actually represent an error


aashir21 (approved):
👍🏻

PR #127, opened 2024-12-16
ShutdownHooks not signalling to main processes because shared-state variable was not volatile

techcobweb:

# Why ?
Processes are not stopping correctly. Even when they get a signal to stop.

See issue: [Resource management pod not restarting on failure #2064](https://github.com/galasa-dev/projectmanagement/issues/2064)

- Same coding pattern used in 5 places !
- Solution isn't overly clean, want to see if it makes a difference.
- A signal should allow the service to shut down now
- Seems to work for the API server, Ctrl-C where I could never do that before.


techcobweb:
Some thoughts:
- This coding pattern could be captured in a common class. As these servers all seem to do the same thing.
- Using synchronised methods via an interface would allow the scope of the shutdown flags to be solely within the server code class, rather than some weird private shutdown hook class.
- No idea if this solves the actual problem which was reported ?

eamansour (approved):
The initial changes seem OK. Of course, there's a lot more to improve on here regarding duplicate code, private classes updating the shutdown flags, unit tests, and cleaning up resources properly. Makes sense to take small steps to burn down these issues, so approved.

PR #130, opened 2024-12-17
initial attempt at runs cleanup logic for runs in DSS with no heartbeat at all

techcobweb:

# Why ?
See issue: [Resource Management should handle test runs without a heartbeat better #2065](https://github.com/galasa-dev/projectmanagement/issues/2065)

If a test run exists in the DSS with no heartbeat time set at all, then it currently gets ignored. Maybe it will get a heartbeat assigned soon, or maybe... it never does, and becomes debris in the dss forever.

This change intends to detect these situations, but keeping tabs on such runs which may turn into debris if enough time elapses, and cleaning them up if they are still there with no heartbeat after that time.

Also, the cache which keeps tabs on these runs needs to forget the run ever existed if it gets a heartbeat assigned once it was being watched...


modules/framework/galasa-parent/dev.galasa.framework/build.gradle
```diff
@@ -20,6 +21,23 @@ dependencies {
...
+    testFixturesImplementation 'commons-io:commons-io'
+    testFixturesImplementation 'org.apache.felix:org.apache.felix.bundlerepository'
+    testFixturesImplementation 'dev.galasa:dev.galasa.wrapping.gson'
+    testFixturesImplementation 'org.apache.bcel:bcel'
+    testFixturesImplementation 'org.apache.commons:commons-lang3'
+    testFixturesImplementation project (':dev.galasa.framework.maven.repository.spi')
+    testFixturesImplementation 'org.apache.logging.log4j:log4j-api'
+    testFixturesImplementation 'org.apache.logging.log4j:log4j-core'
+    testFixturesImplementation 'org.apache.logging.log4j:log4j-slf4j-impl'
+    testFixturesImplementation 'org.yaml:snakeyaml'
```
eamansour: Might be worth checking if we can remove any unnecessary dependencies for these test fixtures

eamansour (approved):
One minor comment regarding dependencies but overall looks good!

modules/framework/galasa-parent/dev.galasa.framework/build.gradle
techcobweb: Yea. Some of them have gone now.

PR #139, opened 2024-12-30
querying a list of roles should return a list of complete role information not just the role metadata

techcobweb:

# Why ?
Part of the [rbac foundation story](https://github.com/galasa-dev/projectmanagement/issues/2071)

- Instead of returning a list of metadata when you query a list of roles, you get all the role information.

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
+++ b/modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
@@ -1938,7 +1938,10 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/RBACRole'
+                type: array
+                description: A list of the roles in the system.
+                items: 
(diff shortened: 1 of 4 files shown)


modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -1938,7 +1938,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/RBACRolesSummary'
+                $ref: '#/components/schemas/RBACRole'
```
eamansour: The success response should return a list/array of roles (currently, it looks like it'll return one role)

eamansour (approved):

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
techcobweb: right. oops.

PR #144, opened 2025-01-02
Iss2085 - Add extra checks in workflows for Forks

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2085

- If the workflow is being run by a fork (so not being run by galasa-dev), skip jobs that are internal to the Galasa team's CI/CD processes such as calling other workflows, recycling the ArgoCD apps that deploy the download sites, and reporting failures into Slack.
- The workflow checks that the required secrets are configured before proceeding, so it doesn't begin the build workflow then fail down the line if a secret isn't set. This means that owners of forks are notified right away of what they need configured in their forked repo to run the build workflows.


.github/templates/settings.xml
```diff
@@ -0,0 +1,42 @@
+<!-- This is a template to show how the GitHub Actions secret MAVEN_SETTINGS_XML should look -->
```
eamansour, line 1: Would it be worth explicitly telling users to not edit this file to prevent them from checking in actual gpg keys and passphrases?

eamansour (approved):

.github/templates/settings.xml
jadecarino: Yes great point - I'll add that in now

PR #148, opened 2025-01-06
removed compat lib. Moved classes so they don't clash with maven core class names

techcobweb:

# Why ?
See story maven-compat 3.6.2 needs upgrading
[#2091](https://github.com/galasa-dev/projectmanagement/issues/2091)

- Removed maven compat as a dependency for the maven plugins (So maven v2 is no longer going to work). Given maven 3.0 was issued in 2010 I don't see this as a problem.
- Moved all the implementation classes down a level, so they don't clash with maven-core classes


techcobweb:
Tested on the CLI-generated projects and it all seems to be working still.

jadecarino (approved):
Looks good, approved

PR #150, opened 2025-01-08
mcobbett users have a role

techcobweb:
# Why ?

See issue https://github.com/galasa-dev/projectmanagement/issues/2071

- [x] user record has a role-id field
- [x] Added user role field to user records
- [x] New users get created with a default role
- [x] If there is no user role set in the user, return the default one
- [x] Synthetic role is returned in the user data structure

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.api.rbac/src/main/java/dev/galasa/framework/api/rbac/RoleTransform.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework.api.rbac/src/main/java/dev/galasa/framework/api/rbac/RoleTransform.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
 package dev.galasa.framework.api.rbac;
 
 import java.util.ArrayList;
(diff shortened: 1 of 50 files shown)


modules/framework/galasa-parent/dev.galasa.framework.api.rbac/src/main/java/dev/galasa/framework/api/rbac/RoleTransform.java
```diff
@@ -1,4 +1,4 @@
-package dev.galasa.framework.api.rbac.internal.routes;
+package dev.galasa.framework.api.rbac;
```
eamansour, line 6: Is this file missing the copyright statement?

modules/framework/galasa-parent/dev.galasa.framework.api.users/src/main/java/dev/galasa/framework/api/users/internal/routes/BeanTransformer.java
```diff
@@ -62,9 +70,54 @@ public UserData convertUserToUserBean(IUser userIn) {
...
     }
 
+    private String getRoleId( IUser userIn ) throws InternalServletException {
+        String roleId = userIn.getRoleId();
+        if (roleId==null || roleId.trim().equals("")) {
+            try {
+                roleId = rbacService.getDefaultRoleId();
+            } catch( RBACException ex) {
+                ServletError err = new ServletError(GAL5086_FAILED_TO_GET_DEFAULT_ROLE);
+                throw new InternalServletException(err, 500, ex);
```
eamansour: Really minor - Could use `HttpServletResponse.SC_INTERNAL_SERVER_ERROR` instead of `500` here

eamansour (approved):
A couple of nitpicks but looks good!

modules/framework/galasa-parent/dev.galasa.framework.api.users/src/main/java/dev/galasa/framework/api/users/internal/routes/BeanTransformer.java
techcobweb: fixed.

PR #162, opened 2025-01-22
Cache user-to-actions entries in the DSS

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2071

## Changes
- Added caching of users-to-actions mappings via a `dss.rbac.<loginId>.actions` DSS property
  -  The property's value is a comma-separated list of action IDs, for example: `GENERAL_API_ACCESS,CPS_PROPERTIES_SET,SECRETS_GET_UNREDACTED_VALUES`
  - These cached properties are stored in the DSS for 24 hours before they expire

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/rbac/CacheRBACImpl.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/rbac/CacheRBACImpl.java
@@ -110,7 +110,7 @@ public class CacheRBACImpl implements CacheRBAC {
 
     private String getUserActionsPropertyKey(String loginId) {
         // The users-to-actions DSS property is in the form:
-        // dss.rbac.user-<loginId>.actions = <comma-separated action IDs>
+        // dss.rbac.user.<loginId>.actions = <comma-separated action IDs>
         return USER_PROPERTY_PREFIX + loginId + ACTIONS_PROPERTY_SUFFIX;
     }
 }
--- a/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/rbac/CacheRBACImpl.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/rbac/CacheRBACImpl.java
@@ -5,9 +5,8 @@
  */
 package dev.galasa.framework.internal.rbac;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
+import java.util.HashSet;
+import java.util.Set;
 
 import dev.galasa.framework.spi.DynamicStatusStoreException;
 import dev.galasa.framework.spi.IDynamicStatusStoreService;
@@ -24,6 +23,7 @@ public class CacheRBACImpl implements CacheRBAC {
     // Only keep users-to-actions entries in the cache for 24 hours
     private static final long CACHED_ACTIONS_TIME_TO_LIVE_SECS = 24 * 60 * 60;
 
+    private static final String USER_PROPERTY_PREFIX = "user.";
     private static final String ACTIONS_PROPERTY_SUFFIX = ".actions";
 
     private IDynamicStatusStoreService dssService;
@@ -41,12 +41,10 @@ public class CacheRBACImpl implements CacheRBAC {
     }
 
     @Override
-    public synchronized void addUser(String loginId, List<String> actionIds) throws RBACException {
+    public synchronized void addUser(String loginId, Set<String> actionIds) throws RBACException {
         try {
-            // Create a property in the DSS of the form:
-            // dss.rbac.<loginId>.actions = <comma-separated action IDs>
             String commaSeparatedActionIds = String.join(",", actionIds);
-            String actionsKey = loginId + ACTIONS_PROPERTY_SUFFIX;
+            String actionsKey = getUserActionsPropertyKey(loginId);
             dssService.put(actionsKey, commaSeparatedActionIds, CACHED_ACTIONS_TIME_TO_LIVE_SECS);
         } catch (DynamicStatusStoreException e) {
             throw new RBACException("Failed to cache user actions", e);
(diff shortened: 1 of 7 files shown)


modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/rbac/CacheRBACImpl.java
```diff
@@ -17,37 +21,76 @@
 
 public class CacheRBACImpl implements CacheRBAC {
 
+    // Only keep users-to-actions entries in the cache for 24 hours
+    private static final long CACHED_ACTIONS_TIME_TO_LIVE_SECS = 24 * 60 * 60;
```
eamansour, line 24: Is 1 day too short? Does this need to be 1 week or longer?

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/rbac/CacheRBACImpl.java
```diff
@@ -17,37 +21,76 @@
...
     }
 
     @Override
     public synchronized void addUser(String loginId, List<String> actionIds) throws RBACException {
-        // Do nothing for now...
+        try {
+            // Create a property in the DSS of the form:
+            // dss.rbac.<loginId>.actions = <comma-separated action IDs>
+            String commaSeparatedActionIds = String.join(",", actionIds);
+            String actionsKey = loginId + ACTIONS_PROPERTY_SUFFIX;
```
eamansour: Is it a good idea to use the login ID in the DSS property's key? Could we use the user number? How would that change things?

We currently grab the username from the auth header in HTTP requests and check access based on the login ID. If we were to use the user number, we would need to grab the user record from the auth store, which is no good.

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/rbac/CacheRBACImpl.java
techcobweb: Think that's fine

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/rbac/CacheRBACImpl.java
techcobweb: ok

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/rbac/CacheRBACImpl.java
```diff
@@ -17,37 +21,72 @@
...
-        IUser user = getUserFromAuthStore(loginId);
-        String userRoleId = user.getRoleId();
-        if (userRoleId == null) {
-            userRoleId = rbacService.getDefaultRoleId();
-            user.setRoleId(userRoleId);
-        }
-        Role userRole = rbacService.getRoleById(userRoleId);
+        boolean isActionPermitted = false;
+        try {
+            String userActionsKey = loginId + ACTIONS_PROPERTY_SUFFIX;
```
techcobweb: Slightly (not much) worried that this will clash with anything else in the dss.
Might be better to prefix it with "user-" or something ?
Not 100% sure whether I should worry about that given we have the suffix also.

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/rbac/CacheRBACImpl.java
```diff
@@ -17,37 +21,72 @@
...
+            String commaSeparatedUserActions = dssService.get(userActionsKey);
+
+            List<String> userActions = new ArrayList<>();
```
techcobweb: Use Set<String> and HashSet<String> as the contains() lookup is loads quicker.

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/rbac/CacheRBACImpl.java
```diff
@@ -17,37 +21,72 @@
...
+            throw new RBACException("Error occurred when accessing the DSS", e);
+        }
         return isActionPermitted;
     }
 
     @Override
     public synchronized void invalidateUser(String loginId) throws RBACException {
-        // Do nothing for now...
+        try {
+            String userActionsKey = loginId + ACTIONS_PROPERTY_SUFFIX;
```
techcobweb: duplicate code calculating the dss key ? Logic would be better only in one place.

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/rbac/CacheRBACImpl.java
eamansour: Added a `user.` prefix in https://github.com/galasa-dev/galasa/pull/162/commits/011d27b7bf48c4812b532fdbf9eb7eb2c9a7be3c, making the property `dss.rbac.user.<loginId>.actions = <actions>`

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/rbac/CacheRBACImpl.java
eamansour: Done in https://github.com/galasa-dev/galasa/pull/162/commits/011d27b7bf48c4812b532fdbf9eb7eb2c9a7be3c

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/rbac/CacheRBACImpl.java
eamansour: Done in https://github.com/galasa-dev/galasa/pull/162/commits/011d27b7bf48c4812b532fdbf9eb7eb2c9a7be3c

techcobweb (approved):

PR #164, opened 2025-01-23
Support customizable VTAM logon strings (#2109)

ideeley:
Adds the ability for users to specify a custom logon string template for a given z/OS image.
New property is zos.image.[TAG].vtam.logon, default value is "LOGON APPLID({0})" to preserve existing behaviour.

Code changes are:
- New class ImageVtamLogon to extract the new property, setting the default behaviour if not found.
- New class TestImageVtamLogon to unit test ImageVtamLogon.
- New method getVtamLogonString in ZosBaseImageImpl to return the fully formatted logon string given a supplied applid (method signature in IZosImage)
- Updates to CicstsDefaultLogonProvider to use the above.
- Zos3270IVT updated to use the above (if it ever gets uncommented).

No additional tests added since existing test classes all have their contents commented out.

Manual testing:
- Run galasa test that logs on to CICS region, new property not specified - logs on using the correct default.
- Repeat with the new property specified with the default - logs on using the default
- Repeat with the new property specified with a different pattern - attempts log on with the expected custom logon string (logon fails since my test system only accepts the default logon command - but that was expected).

Local build is clean (including run of the new jUnits).


techcobweb (approved):

techcobweb:
This PR is implements story https://github.com/galasa-dev/projectmanagement/issues/2109

PR #166, opened 2025-01-27
GALASA_DEFAULT_USER_ROLE is used to select the default role for the user.

techcobweb:

# Why ?

This is part of the basic RBAC story: [Minimal Admin/Non-admin RBAC #2071](https://github.com/galasa-dev/projectmanagement/issues/2071)

The user can set the GALASA_DEFAULT_USER_ROLE environment variable to be `tester` `deactivated` or `admin`. Default is `tester`.

When set, a newly created user will have this role.

Even if it is set, any users which exist in the system prior to this code change will have a
`synthetic` role of `admin` set (even if their role is not in their user record).
This seemed the fairest thing to do, as all users today are essentially allowed to do anything. Then systems admins can down-grade people's permissions as they wish.


modules/managers/galasa-managers-parent/galasa-managers-comms-parent/dev.galasa.http.manager/src/main/java/dev/galasa/http/HttpClientResponse.java
```diff
@@ -273,32 +273,33 @@ public static HttpClientResponse<JsonObject> jsonResponse(CloseableHttpResponse
...
                         }
                     } else {
-                        logger.info("Did not attempt to parse JSON from the following: " + sResponse);
+                        EntityUtils.consume(httpResponse.getEntity());
                     }
-                } else {
-                    EntityUtils.consume(httpResponse.getEntity());
                 }
+            } finally {
+                httpResponse.close();
```
techcobweb, line 300: HTTP responses were not being closed if an exception occurred.

modules/maven/galasa-maven-plugin/pom.xml
```diff
@@ -189,8 +189,8 @@
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-compiler-plugin</artifactId>
 				<configuration>
-					<source>11</source>
-					<target>11</target>
+					<source>17</source>
```
techcobweb: I was getting a bad error on class loading where the Java 11 code format wasn't working here. Changing it up to Java 17 solved my problem. Not sure if this should be submitted or not. Depends if others are finding the same thing when they build everything from scratch.

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -206,12 +206,13 @@ public List<IUser> getAllUsers() throws AuthStoreException {
     @Override
     public void createUser(String loginId, String clientName, String roleId) throws AuthStoreException {
 
+        logger.info("createUser: Creating user loginId"+loginId+" clientName:"+clientName+" roleId:"+roleId );
```
jadecarino: Colon missing here in the logger message might make things hard to read if the loginId starts with letters potentially

modules/managers/galasa-managers-parent/galasa-managers-comms-parent/dev.galasa.http.manager/src/main/java/dev/galasa/http/HttpClientResponse.java
jadecarino: Good catch

modules/maven/galasa-maven-plugin/pom.xml
jadecarino: Makes sense. I'm not sure either on the delivery of this though. Will it break users with older Java runtimes?

jadecarino: Personally I've not had any issues building everything from scratch but can't say for others.

jadecarino (approved):
One minor comment and a response to your comment about the Java source/target versions which might need looking into before merging. Otherwise, the code changes all look good. Approved!

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
techcobweb: fixed

modules/maven/galasa-maven-plugin/pom.xml
techcobweb: Reverted it now I've upgraded to mvn 3.9.0

PR #167, opened 2025-01-27
Workflow improvements to better support forks

jadecarino:
Changes to better support forks:
 - Add `workflow_dispatch` option for Main Build Orchestrator so it can be ran with a button, as first time contributors setting up their forks will need to run an initial main build to populate their artifact cache and packages.
- Change `WRITE_GITHUB_PACKAGES_USERNAME` to be a workflow variable rather than a repository secret, as it's not confidential information.
- Remove need for `MAVEN_SETTINGS_XML` to be set and instead provide a settings.xml file and then insert the GPG key name and passphrase in at build time with environment variables.
- The check-required-secrets-configured workflow now notifies the user to check the output of the "Report" job to find more information on what they are missing.

Other changes:
- Fix the get-changed-modules-pull-request script as it was hard coded to search Pull Requests for the galasa-dev/galasa repo, now it correctly checks Pull Requests for the HEAD repo of the Pull Request based on the `github` context.
- Check for required secrets and variables before proceeding with the Release Build Orchestrator.
- Do not proceed with Release Build Orchestrator if it is triggered by a fork.


.github/common-files/settings.xml
```diff
@@ -12,9 +10,8 @@
                 <activeByDefault>true</activeByDefault>
             </activation>
             <properties>
-                <!-- Substitute your GPG key information here: -->
-                <gpg.keyname>{{ yourGpgKeyName }}</gpg.keyname>
-                <gpg.passphrase>{{ yourGpgPassphrase }}</gpg.passphrase>
+                <gpg.keyname>${gpg.key.id}</gpg.keyname>
```
techcobweb, line 13: loads better

techcobweb (approved):
it all makes sense. the settings.xml change is really good. one less thing to set.

.github/common-files/settings.xml
jadecarino: Thanks for the review :)

PR #174, opened 2025-01-30
Run script gets its OBR version dynamically, same as the galasa version.

techcobweb:

# Why ?

I got worried that every time I was running the `run-locally.sh` script that it was launching using the wrong OBR version, as the OBR version was hard-coded into the script.

I changed the script to dig out the version from the `build.gradle` file, which will be the same as the OBR version now that everything uses the same versioning scheme.


aashir21 (approved):
👍🏻

PR #178, opened 2025-01-30
stop user updating their own role

techcobweb:

# Why ?
Doing more testing of the RBAC functionality it became clear that no user should be able to change their own role.

This means that the last admin user can't set themselves to a non-admin, making it less likely that we will see admin lock-out... a system where there are no admins.

- [x] Code the check.
- [x] Add a unit test.

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.api.users/src/main/java/dev/galasa/framework/api/users/internal/routes/UserRoute.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework.api.users/src/main/java/dev/galasa/framework/api/users/internal/routes/UserRoute.java
@@ -11,6 +11,12 @@ import java.util.regex.Pattern;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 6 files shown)


modules/framework/galasa-parent/dev.galasa.framework.api.users/src/main/java/dev/galasa/framework/api/users/internal/routes/UserRoute.java
```diff
@@ -83,7 +91,8 @@ public HttpServletResponse handlePutRequest(
         HttpServletResponse response
     ) throws FrameworkException, IOException {
 
-        validateActionPermitted(BuiltInAction.USER_EDIT_OTHER, requestContext.getUsername());
+        String requestingUserLoginId = requestContext.getUsername();
+        validateActionPermitted(BuiltInAction.USER_EDIT_OTHER, requestingUserLoginId);
```
jadecarino, line 95: Prefer this :)

jadecarino (approved):
Good stuff approved

PR #180, opened 2025-02-03
Introduce FrameworkType and set Run name of Resource Management

### LINKED ISSUE #2124: Ability to access managers during Resource Management
## Story
As a Infrastructure admin, I want resource management to be able to clean up resource on systems not just the DSS, so that the infrastructure is cleaned up correctly.

## Background
Currently resource management is unable to access systems as part of its execution.  For example we have a custom manager that creates Datasets on the z/OS machine per test.  These are logged into the DSS and are cleaned up if the test runs cleanly.  If for any reason the pod fails within the ecosystem and so there is no clean up done these datasets are left.  The resource manager is running in the background and can clean up the DSS but there is no way to access the zOS machine itself to perform the cleanup.

We have a couple of other managers that have this same issue as well leaving jobs running using pooled resource ids.  If the same Id is selected again by another test then that test will fail because the resource id is already in use (but not in use according to the DSS).

I've been experimenting to see whether it would be possible to get this to work from without our resource manager and with a lot of code it is just about possible to start all the bundles up.  However there are a number of issues with this:

- Currently the RAS isn't setup correctly - Even with passing in a RAS setting it's not setup correctly because everything assumes there is a Galasa Test involved (including run names).  (See below)
- All managers expect to be able to find `@` annotations of things they want to create, but in the resource management environment there are no annotations because there is no test class.
- Loading of the bundles per resource manager requires a lot of code and it would be better to have all this code within the resource management code instead.  If it's required we attempt to start bundles just like happens within the main TestBundles.

### RAS
The RAS one is interesting because after abit of experimenting I've managed to get the resource management to have a RAS by setting the runName (note this was running locally with a Directory RAS, see [DirectoryArchiveStoreService](https://github.com/galasa-dev/galasa/blob/f9097c85d093900ef27fe9504ac598695775330e/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/ras/directory/DirectoryResultArchiveStoreService.java#L72).  Looking at CouchDB implementation I don't think it would have the same issues.  However I am worried what setting this will mean in the long term but more experimenting will be needed.  

## Tasks
This is an attempt at breaking things down into manageable parts.  There are some fundamental changes within the framework that I think are required (note this might not be everything that is required!)

- [x] Update FrameworkInitialisation.java to initialise the framework differently depending on whether the resource manager, a test pod or something else starts the framework up. Added FrameworkInitialisationStrategy to cover this. See https://github.com/galasa-dev/galasa/commit/2d2a33f39956b74e215711013b74fb55c0ba9f18
- [x] Update FrameworkInitialisation.java to use the enum to decide whether to set the testRunName or not and what to set it as.  Default would be no RunName is set, if Test = current behaviour, if ResourcManagement = "resourceManagement".  We may need to use this in other places to handle this better. See https://github.com/galasa-dev/galasa/commit/2d2a33f39956b74e215711013b74fb55c0ba9f18

- [ ] Implement AbstractResourceManager that uses could extend (this would contain the code for starting up bundles, similar to how AbstractManager is done).  This would allow calls like youAreDependent to work similar to AbstractManager.
- [ ] Add to the z/OS Manager a new SPI interface called IZosManagerResourceSpi that the ZosManagerImpl implements.  Allowing a calling manager to get to the IZosImage to perform actions.  For example something like:

```
public interface IZosManagerResourceSpi
{
    public IZosImage getImageResource(String imageId) throws ZosManagerException;
}
```
- [ ] Create test manager that leaves behind a resource, does not get cleaned up if a test is cancelled, to have resources to clean to test changes are working

adamcoulthard:
fixes galasa-dev/projectmanagement#2124

This is the first part of the Resource Management changes.  To allow RAS directory pieces to start to work correctly this introduces the new internal FrameworkType enum at the moment that are the following enums:

- test
- resourceManagement
- ecosystem
- launcher
- controller
- server

Some of these could be expanded but at the moment the only two that actually mean anything are test and resourceManager.  When the framework is initialised it uses these to decide what to do with the testRunName which is either set to the next run number if a test or "resourceManagement" if its the resourceManagement, otherwise no RunName is set.

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/FrameworkInitialisation.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/FrameworkInitialisation.java
@@ -38,22 +38,13 @@ public class FrameworkInitialisation implements IFrameworkInitialisation {
     private String galasaHome;
 
     private static final GalasaGson gson = new GalasaGson();
-    
-    public FrameworkInitialisation(
-        Properties bootstrapProperties, 
-        Properties overrideProperties
-    ) throws URISyntaxException, InvalidSyntaxException, FrameworkException {
-        this(bootstrapProperties, overrideProperties, false, null, 
-        getBundleContext() , new FileSystem() , new SystemEnvironment());
-    }
-
 
     public FrameworkInitialisation(
         Properties bootstrapProperties, 
         Properties overrideProperties, 
-        boolean testrun
+        FrameworkType type
     ) throws URISyntaxException, InvalidSyntaxException, FrameworkException {
-        this(bootstrapProperties, overrideProperties, testrun, null, 
+        this(bootstrapProperties, overrideProperties, type, null, 
         getBundleContext() , new FileSystem(), new SystemEnvironment());
     }
 
@@ -61,17 +52,17 @@ public class FrameworkInitialisation implements IFrameworkInitialisation {
     public FrameworkInitialisation(
         Properties bootstrapProperties, 
         Properties overrideProperties,
-        boolean testrun,
+        FrameworkType type,
         Log initLogger
     ) throws URISyntaxException, InvalidSyntaxException, FrameworkException {
-        this(bootstrapProperties, overrideProperties, testrun, initLogger, 
+        this(bootstrapProperties, overrideProperties, type, initLogger, 
         getBundleContext(), new FileSystem(), new SystemEnvironment());
     }
 
     public FrameworkInitialisation(
         Properties bootstrapProperties, 
         Properties overrideProperties, 
-        boolean testrun,
+        FrameworkType type,
         Log initLogger,
         BundleContext bundleContext , 
         IFileSystem fileSystem,
@@ -105,7 +96,7 @@ public class FrameworkInitialisation implements IFrameworkInitialisation {
         this.framework.setFrameworkProperties(overrideProperties);
 
         // *** If this is a test run, then we need to install the log4j capture routine
-        if (testrun) {
+        if (type == FrameworkType.test) {
             this.framework.installLogCapture();
         }
 
@@ -117,13 +108,23 @@ public class FrameworkInitialisation implements IFrameworkInitialisation {
             this.logger, overrideProperties, this.cpsFramework, this.fileSystem);
         this.dssFramework = initialiseDynamicStatusStore(logger,bundleContext);
 
-        if (testrun) {
-            // *** Is this a test run,
-            // *** Then we need to make sure we have a runname for the RAS. If there isnt
-            // one, we need to allocate one
-            // *** Need the DSS for this as the latest run number number is stored in there
-            String runName = locateRunName(this.cpsFramework);
-            this.framework.setTestRunName(runName);
+        switch(type) {
+            case test: {
+                // *** Is this a test run,
+                // *** Then we need to make sure we have a runname for the RAS. If there isnt
+                // one, we need to allocate one
+                // *** Need the DSS for this as the latest run number number is stored in there
+                String runName = locateRunName(this.cpsFramework);
+                this.framework.setTestRunName(runName);
+            } break;
+
+            case resourceManagement: {
+                this.framework.setTestRunName("resourceManagement");
+            } break;
+
+            default: {
+                // The default is we don't set the TestRunName
+            } break;
         }
 
         this.uriResultArchiveStores = createUriResultArchiveStores(overrideProperties, this.cpsFramework);
@@ -146,7 +147,7 @@ public class FrameworkInitialisation implements IFrameworkInitialisation {
 
         // *** If this is a test run, add the overrides from the run dss properties to
         // these overrides
-        if (testrun) {
+        if (type == FrameworkType.test) {
             loadOverridePropertiesFromDss(overrideProperties);
         }
     }
--- a/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/FrameworkType.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/FrameworkType.java
@@ -1,0 +1,10 @@
+/*
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package dev.galasa.framework;
+
+public enum FrameworkType {
+    test, resourceManagement, ecosystem, launcher, controller, server
+}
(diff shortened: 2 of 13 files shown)


techcobweb:
So far in Galasa development we've been really trying hard to not make the framework know how it's being used.
Not looked at the code yet, but this is something we've been actively trying to avoid.
Any variance we've added a plug-point, and the different configurations inject a different implementation.
eg: Ras using etcd or Ras using a file.
Will look at the code and get back to you...


...some time passes...

OK. Looked at the code. It's clear we didn't follow our own advice on that one (soo inconsistent), so what you are doing isn't a huge stretch beyond what's already there.

And this PR isn't disruptive I believe.

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/FrameworkInitialisation.java
```diff
@@ -117,13 +108,23 @@ public FrameworkInitialisation(
...
+                // *** Is this a test run,
+                // *** Then we need to make sure we have a runname for the RAS. If there isnt
+                // one, we need to allocate one
+                // *** Need the DSS for this as the latest run number number is stored in there
+                String runName = locateRunName(this.cpsFramework);
+                this.framework.setTestRunName(runName);
+            } break;
+
+            case resourceManagement: {
+                this.framework.setTestRunName("resourceManagement");
```
techcobweb, line 122: This "resourceManagement" should be a symbolic literal from somewhere.

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/FrameworkInitialisation.java
```diff
@@ -117,13 +108,23 @@ public FrameworkInitialisation(
...
         this.dssFramework = initialiseDynamicStatusStore(logger,bundleContext);
 
-        if (testrun) {
-            // *** Is this a test run,
-            // *** Then we need to make sure we have a runname for the RAS. If there isnt
-            // one, we need to allocate one
-            // *** Need the DSS for this as the latest run number number is stored in there
-            String runName = locateRunName(this.cpsFramework);
-            this.framework.setTestRunName(runName);
+        switch(type) {
```
techcobweb, line 111: Might be worth putting this into a separate method so the overall method doesn't get any bigger.

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/FrameworkType.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package dev.galasa.framework;
+
+public enum FrameworkType {
+    test, resourceManagement, ecosystem, launcher, controller, server
```
techcobweb, line 9: TEST, RESOURCE_MANAGEMENT, ECOSYSTEM, LAUNCHER, CONTROLLER, SERVER - I always thought enums should be upper-case names. I see there are differing opinions online about this, but generally in Java I think all constants are upper-case, and enum values are constants.

What do each mean ? eg: What is an ECOSYSTEM type of framework ? Not sure that means anything to me.

Also, trying to move away from the word "ecosystem" in general anyway

What's the difference between a LAUNCHER and SERVER ?...etc.

As you don't have a need for some of these types, maybe have TEST, RESOURCE_MANAGEMENT, OTHER ? until we do need the distinctions ?

techcobweb (changes requested):
I see what you're doing. The changes don't look disruptive. Some minor points raised in the comments.

techcobweb:
In some ways I wonder whether we should just pass in a TestNameAllocator(), and OverridePropertiesLoader() interface/class rather than do this, to drive the differences in behaviour. That would make for fewer branches in the code, and fewer unit tests to write generally.

Or (better still) have a FrameworkTypeStrategy interface, which gets passed in, which hides the differences between the different 'modes' of running.
eg:
```java
public interface FrameworkInitialisationStrategy {

  /** @returns null if no testcase name needs to be set
  String getTestcaseName(cps);

  void startLoggingCapture(framework);

  Properties loadOverridePropertiesFromDss(dss);
 
  ...etc.
}
```

We could just have a Test implementation, and a ResourceManager implementation, and a default implementation.

We could also extract some code out of the initialiser class which is specific to Test initialisation, or resource manager initialisation..etc. eg: locateRunName()

Is it important for your future plans that the type lives beyond the initialisation of the framework, to provide different behaviour? Like what ?...

techcobweb:
Also, unit testing... any scope for some unit tests at all here ?

If unit testing can be done without relying on any framework, just straight Java old-school way, then we lessen our dependencies also.

The unit testing gets easier if we inject a strategy, and unit test that each implementation of the strategy does what it's supposed to independently of the base framework code.

techcobweb:
@adamcoulthard Thanks for the PR to look at. Comments above. Happy to talk about anything I've written there.

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/FrameworkInitialisation.java
adamcoulthard: Any ideas where might be best to place this?  Is there a standard location for these things?  For the time being I could just add it to the FrameworkInitialisation.java.

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/FrameworkType.java
adamcoulthard: I did wonder about not doing them all, happy to change to these.

adamcoulthard:
@techcobweb there is no reason for anything to live on.  The only reason really for this change is to allow me to set the testName for the resourceManager and prior to these changes I couldn't do it.  This allows the managers to be able to initialise their Ras Providers correctly.

I think longer term we probably want to look at the managers and how the RAS stuff works so that we can make a null RAS provider that just "swallows" everything.  But I've not really understood full how all the RAS stuff hangs together and was a lot more nervous about making changes in that area than this simple fix now.  With potential enhancing in the future.

I would probably agree that doing something slight different to all this would make sense but I was worried about over complicating the startup.  Its already difficult to weave though and I thought more indirection would make things worse.

techcobweb:
Would you like me to re-implement these changes in the way I suggest above ?

adamcoulthard:
@techcobweb if you feel it would be better done using the new class then go for it. I’ve not had chance to get back to this and implement your comments.

adamcoulthard:
Closing this PR because its been superceded by one from @techcobweb

PR #181, opened 2025-02-04
If a user has a valid JWT but their user record is deleted, say no permissions instead of internal server error

techcobweb:

# Why ?
See [http 500 instead of FORBIDDEN when trying to access using an invalid token. #2119](https://github.com/galasa-dev/projectmanagement/issues/2119)

- When the code checks to see if this loginId has permissions to do something
- (So the caller must already have a valid JWT)
- There is no cache hit for that user's permissions in the dss.
- So the code goes to the auth/users store to look up that user.
- If the user has been deleted/not found, then it's not a 500 error, but a 403 Forbidden one.

- Unit test changed to make sure this is the case.
- [ ] Tested locally by deleting the user then trying to use the users' JWT.


jadecarino (approved):
Looks good

PR #184, opened 2025-02-04
Blank property values are allowed

techcobweb:

# Why ?
See issue [Galasa Resources Apply fails if properties are given an empty value, even though ETCD accepts properties that are empty #2115](https://github.com/galasa-dev/projectmanagement/issues/2115)

# Tasks
- relax some validation checks to allow blanks.
- adjust unit tests to test that blank values are allowed.
- expand some error messages so that they are unique, and capture all the message text in the  correct place so messages could later be translated.

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.api.common/src/test/java/dev/galasa/framework/api/common/resources/TestCPSProperty.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework.api.common/src/test/java/dev/galasa/framework/api/common/resources/TestCPSProperty.java
@@ -243,7 +243,7 @@ public class TestCPSProperty extends BaseServletTest {
     }
 
     @Test
-    public void testCPSPropertyBlankValueIsInvalid() throws InternalServletException{
+    public void testCPSPropertyBlankValueIsValid() throws InternalServletException{
         //Given...
         String namespace = "framework";
         String propertyName = "property";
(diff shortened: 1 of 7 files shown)


modules/framework/galasa-parent/dev.galasa.framework.api.common/src/test/java/dev/galasa/framework/api/common/resources/TestCPSProperty.java
```diff
@@ -243,7 +243,7 @@ public void testCPSPropertyNoNameIsInvalid() throws InternalServletException{
     }
 
     @Test
-    public void testCPSPropertyNoValueIsInvalid() throws InternalServletException{
+    public void testCPSPropertyBlankValueIsInvalid() throws InternalServletException{
```
jadecarino: This should say `testCPSPropertyBlankValueIsValid`

jadecarino (changes requested):
One minor change on the test method name but looks good otherwise

modules/framework/galasa-parent/dev.galasa.framework.api.common/src/test/java/dev/galasa/framework/api/common/resources/TestCPSProperty.java
techcobweb: yea

techcobweb: Fixed

jadecarino (approved):

PR #188, opened 2025-02-11
Iss2101 - Start of contributor's guide

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2101

There is still work I'd like to do on this but thought it makes sense to deliver it as is so it's available for contributors to read now, and expand on it in a later change. I've also corrected some links in the module README.mds that refer to the old archived repositories.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,6 +68,20 @@ To set a repository variable:
 1. GPG_PASSPHRASE: This requires your passphrase for your GPG key in plain text.
 1. WRITE_GITHUB_PACKAGES_TOKEN: This requires your GitHub Personal Access Token with write:packages scope that you want to use to log into GitHub Container Registry.
 
+For the GPG related secrets, you can either use an existing key, or create a new key.
+
+**How to create a new GPG key**:
+1. Generate a new GPG key with `gpg --full-generate-key`
+2. Select the type of key using the following options:
+    * RSA and RSA (default)
+    * 4096 bits long
+    * Select an expiration date
+3. Enter your User ID for the key:
+    * Name: Your full name
+    * Email: Your email address
+    * Comment: N/A
+4. Enter a password, and ensure to remember it or note it down, as you will need it later.
+
 **How to set the GPG_KEYID repository secret**: 
 
 1. Get your GPG key information with `gpg --list-secret-keys --keyid-format=short`. The output will contain something like the block below. In this example, your GPG key ID is `XXXXXXXX`. 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,22 +25,29 @@ Welcome to Galasa! To learn more about contributing to this repository, please r
 
 - Fix typos, improve examples, or enhance explanations.
 
-## Making your first contribution
+## How to make a contribution?
 
-### How to set up a fork of a repository
+### Set up a fork of a repository
 
 1. On GitHub, navigate to the repository.
-1. In the top-right corner of the page, click Fork. (insert screenshot)
+1. In the top-right corner of the page, click Fork.
 1. Select an owner for the forked repository from the dropdown menu under "Owner".
 1. The fork will be named the same as the upstream repository as default. Optionally, to further distinguish your fork, type a name in the "Repository name" field.
 1. Optionally, type a description for your fork in the "Description" field.
 1. Optionally, select "Copy the `main` branch only".
-1. You will need to enable the GitHub Actions workflows by going to the Actions tab and clicking 'I understand my workflows, go ahead and enable them'
-1. You will need to trigger an initial workflow run of the 'Main Build Orchestrator' before making any Pull Requests from a branch to your main branch on your fork. You can do this by going to the Actions tab, clicking 'Main Build Orchestrator', and submitting 'Run workflow'. (todo: expand on this, explain about how the artifact caching works)
+1. Click "Create fork".
 
-### How to configure repository secrets and variables
+### Get your fork ready to start contributing and building
 
-To run certain workflows on your fork, you will need to configure some repository secrets and variables so that the workflows can access them. (todo: expand...)
+The 'galasa' repository is divided into several modules that are stored within the `/modules` directory. Certain modules require artifacts from other modules as they have a dependency on them. The build workflow attempts to download these artifacts either from the same workflow run or from a previously successful completed workflow run. Therefore, before you can open a Pull Request on your forked repository from one branch to another, you will need to run an initial workflow run of the "Main Build Orchestrator".
+
+1. As default, GitHub Actions will be disabled on your fork. You will need to enable the GitHub Actions workflows by going to the Actions tab on your fork and clicking "I understand my workflows, go ahead and enable them".
+1. To run the "Main Build Orchestrator" on your fork, you will need to configure some repository secrets and variables. Follow the section below [Configure repository secrets and variables](#configure-repository-secrets-and-variables).
+1. Trigger an initial workflow run of the "Main Build Orchestrator". You can do this by going to the Actions tab, selecting "Main Build Orchestrator" from the list of workflows on the left (you may need to click "Show more workflows..."), then clicking "Run workflow" on the right.
+
+### Configure repository secrets and variables
+
+To run certain workflows on your fork, you will need to configure some repository secrets and variables so that the workflows can access them. These secrets and variables are used for things such as authenticating to the GitHub Container Registry and supplying Maven and Gradle builds with your GPG key information to sign artifacts.
 
 #### Repository secrets and variables that required configuring
 
@@ -66,7 +73,7 @@ To set a repository variable:
 1. Select 'New repository secret'.
 1. Enter the secret name and value, then click 'Add secret'. Once you have done this, you will not be able to view the secret value again.
 
-**Required repository secrets:**
+###### Required repository secrets:
 
 1. GPG_KEY: This requires your Base 64 encoded GPG key payload.
 1. GPG_KEYID: This requires the ID of your GPG key in plain text.
@@ -75,7 +82,7 @@ To set a repository variable:
 
 For the GPG related secrets, you can either use an existing key, or create a new key.
 
-**How to create a new GPG key**:
+###### How to create a new GPG key:
 1. Generate a new GPG key with `gpg --full-generate-key`
 2. Select the type of key using the following options:
     * RSA and RSA (default)
@@ -87,7 +94,7 @@ For the GPG related secrets, you can either use an existing key, or create a new
     * Comment: N/A
 4. Enter a password, and ensure to remember it or note it down, as you will need it later.
 
-**How to set the GPG_KEYID repository secret**: 
+###### How to set the GPG_KEYID repository secret: 
 
 1. Get your GPG key information with `gpg --list-secret-keys --keyid-format=short`. The output will contain something like the block below. In this example, your GPG key ID is `XXXXXXXX`. 
 ```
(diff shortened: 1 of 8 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,166 @@
...
+1. WRITE_GITHUB_PACKAGES_TOKEN: This requires your GitHub Personal Access Token with write:packages scope that you want to use to log into GitHub Container Registry.
+
+For the GPG related secrets, you can either use an existing key, or create a new key.
+
+###### How to create a new GPG key:
+1. Generate a new GPG key with `gpg --full-generate-key`
+2. Select the type of key using the following options:
+    * RSA and RSA (default)
+    * 4096 bits long
+    * Select an expiration date
```
techcobweb: For example, 1y ?

CONTRIBUTING.md
```diff
@@ -0,0 +1,166 @@
...
+
+1. GPG_KEY: This requires your Base 64 encoded GPG key payload.
+1. GPG_KEYID: This requires the ID of your GPG key in plain text.
+1. GPG_PASSPHRASE: This requires your passphrase for your GPG key in plain text.
+1. WRITE_GITHUB_PACKAGES_TOKEN: This requires your GitHub Personal Access Token with write:packages scope that you want to use to log into GitHub Container Registry.
+
+For the GPG related secrets, you can either use an existing key, or create a new key.
+
+###### How to create a new GPG key:
+1. Generate a new GPG key with `gpg --full-generate-key`
```
techcobweb, line 105: Note: The Galasa build requires a GNU Privacy Guard (GPG) key to sign built artifacts so it can be proven who built them. Github also requires a GPG key if you plan on signing commits (see <link> how to set that up). These two GPG keys can be the same if you wish, but they can be separate and different.

Here we explain how to create a new key from scratch.

First you will need to make sure the `gpg` command is installed on your system.

CONTRIBUTING.md
```diff
@@ -0,0 +1,166 @@
...
+2. Select the type of key using the following options:
+    * RSA and RSA (default)
+    * 4096 bits long
+    * Select an expiration date
+3. Enter your User ID for the key:
+    * Name: Your full name
+    * Email: Your email address
+    * Comment: N/A
+4. Enter a password, and ensure to remember it or note it down, as you will need it later.
```
techcobweb: Enter a password, and ensure to remember it or note it down (in a secure place), as you will need it later.

techcobweb: Also, it should be a passphrase I think. As when you have to supply it they ask for the passphrase. Worth checking.

CONTRIBUTING.md
```diff
@@ -0,0 +1,166 @@
...
+```
+sec   rsa4096/XXXXXXXX 2023-05-22 [SC] [expires: 2025-05-21]
+      123456789101112131415161718192021222324252627282930
+```
+2. Create a repository secret called `GPG_KEYID` and add the value of the GPG key ID in plain text.
+
+
+###### How to set the GPG_KEY repository secret: 
+
+1. Use your GPG key ID from above to get your GPG key payload in a Base 64 encoded format with `gpg --export-secret-keys XXXXXXXX | base64`. Ensure it is on one line.
```
techcobweb: you will be prompted for the passphrase that you noted earlier.

CONTRIBUTING.md
```diff
@@ -0,0 +1,166 @@
...
+1. Add the output to your repository secrets for GPG_KEY.
+
+###### How to set the GPG_PASSPHRASE repository secret:
```
techcobweb: Move this as the first step, so they don't have to remember the passphrase for so long?

CONTRIBUTING.md
```diff
@@ -0,0 +1,166 @@
...
+1. Enter the secret name and value, then click 'Add secret'. Once you have done this, you will not be able to view the secret value again.
+
+###### Required repository secrets:
+
+1. GPG_KEY: This requires your Base 64 encoded GPG key payload.
+1. GPG_KEYID: This requires the ID of your GPG key in plain text.
+1. GPG_PASSPHRASE: This requires your passphrase for your GPG key in plain text.
+1. WRITE_GITHUB_PACKAGES_TOKEN: This requires your GitHub Personal Access Token with write:packages scope that you want to use to log into GitHub Container Registry.
+
+For the GPG related secrets, you can either use an existing key, or create a new key.
```
techcobweb, line 96: Note: The Github web interface allows the setting of secrets, but you can't read them back, so be very careful what you set them to so you avoid some painful problem diagnosis steps later.

CONTRIBUTING.md
```diff
@@ -0,0 +1,166 @@
...
+1. Add it to your repository secrets for GPG_PASSPHRASE.
+
+###### How to set the WRITE_GITHUB_PACKAGES_TOKEN repository secret: 
+
+You will first need to create a new GitHub Personal access token (classic) or Fine-grained personal access token.
+
+To create a GitHub Personal access token (classic):
+1. Go to the Settings for your GitHub account > Developer settings > Personal access tokens > Tokens (classic). 
+1. Select Generate new token (classic).
+1. Select write:packages access and give the token a name.
```
techcobweb: 1. Give the token a name, such as `token to build personal forks of Galasa`
2. Select an appropriate expiration time. For example `custom` in a years' time.
3. Select the `write:packages` access
4. Press the `Generate Token` button at the bottom of the dialog.

CONTRIBUTING.md
```diff
@@ -0,0 +1,166 @@
...
+1. The fork will be named the same as the upstream repository as default. Optionally, to further distinguish your fork, type a name in the "Repository name" field.
+1. Optionally, type a description for your fork in the "Description" field.
+1. Optionally, select "Copy the `main` branch only".
+1. Click "Create fork".
+
+### Get your fork ready to start contributing and building
+
+The 'galasa' repository is divided into several modules that are stored within the `/modules` directory. Certain modules require artifacts from other modules as they have a dependency on them. The build workflow attempts to download these artifacts either from the same workflow run or from a previously successful completed workflow run. Therefore, before you can open a Pull Request on your forked repository from one branch to another, you will need to run an initial workflow run of the "Main Build Orchestrator".
+
+1. As default, GitHub Actions will be disabled on your fork. You will need to enable the GitHub Actions workflows by going to the Actions tab on your fork and clicking "I understand my workflows, go ahead and enable them".
```
techcobweb, line 46: Put this in a separate ### Section ? Enable github actions

CONTRIBUTING.md
```diff
@@ -0,0 +1,166 @@
...
+1. To run the "Main Build Orchestrator" on your fork, you will need to configure some repository secrets and variables. Follow the section below [Configure repository secrets and variables](#configure-repository-secrets-and-variables).
+1. Trigger an initial workflow run of the "Main Build Orchestrator". You can do this by going to the Actions tab, selecting "Main Build Orchestrator" from the list of workflows on the left (you may need to click "Show more workflows..."), then clicking "Run workflow" on the right.
```
techcobweb: As you can't do this until the secrets are configured, move it to the end after the secrets ? Then you don't have to remember to return to this point to kick it off.
Maybe 

### Manually kick-off the first `Main Build Orchestrator` workflow
To prime build caches with downloaded dependencies, and to check that your variables and secrets are correctly configured, you will need to manually trigger the "Main Build Orchestrator" workflow.

- Navigate to the Actions tab
- Select the "Main Build Orchestrator" workflow from the list of workflows on the left.....etc.

A list of steps is easier to follow than prose.

techcobweb: Also, Give some expectation on how long it will take first time through ?
And how long would we expect it to take after the first time ?

techcobweb: How would I know if it worked ? or failed ?
What to do if it fails ?

techcobweb: Should I be worried about these warnings:
image
(And do we have an issue to remove them? - need to raise it if not I think)

Maybe this is expected, if so lets mention it so people don't panic.

techcobweb: Is `Caching for Gradle actions was disabled - expand for details` normal ?

techcobweb: Maybe we should add some summary information into the workflow steps. https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#example-of-adding-a-job-summary

techcobweb: If the secrets/keys/variables were not set correctly, how far would you expect the build to get before failing ?

techcobweb: If it fails, will it be obvious how to fix it/what to correct ?

techcobweb (changes requested):
I've asked for a few changes, but if you disagree with anything or need clarification as to what I meant.... happy to discuss...etc.

CONTRIBUTING.md
jadecarino: done

CONTRIBUTING.md
jadecarino: done

CONTRIBUTING.md
jadecarino: done

CONTRIBUTING.md
jadecarino: done

CONTRIBUTING.md
jadecarino: done

CONTRIBUTING.md
jadecarino: done

CONTRIBUTING.md
jadecarino: done

CONTRIBUTING.md
jadecarino: thanks for the feedback - I'll move this section to after the secrets section, and improve the readability.

jadecarino: I'm not too sure what those annotations warnings are about, they're not expected, so yes I'll raise a story to remove these

jadecarino: Yes we have unfortunately had to disable caching in the workflows, as caching cached everything even dev.galasa artifacts, so we couldn't pick up code changes to dev.galasa artifacts during builds so the newest code would never be built deployed

jadecarino: I'll raise a story for adding summary information thats a good idea

CONTRIBUTING.md
jadecarino: done

CONTRIBUTING.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Configure repository secrets and variables
+
+To run certain workflows on your fork, you will need to configure some repository secrets and variables so that the workflows can access them. These secrets and variables are used for things such as authenticating to the GitHub Container Registry and supplying Maven and Gradle builds with your GPG key information to sign artifacts.
+
+#### Repository secrets and variables that required configuring
```
techcobweb: 'require' or 'are required' and drop the 'configuring'

CONTRIBUTING.md
```diff
@@ -0,0 +1,166 @@
...
+To set a repository variable:
+1. Navigate to your repository settings.
+1. Select 'Secrets and variables', then 'Actions' from the menu.
+1. Click the 'Variables' tab.
+1. Select 'New repository variable'.
+1. Enter the variable name and value, then click 'Add variable'.
+
+**Required repository variables:**
+
+1. WRITE_GITHUB_PACKAGES_USERNAME: This requires the GitHub username you want to use to log into GitHub Container Registry.
```
techcobweb: Is this checked against anything ? does it have to match anything ?

CONTRIBUTING.md
jadecarino: done

CONTRIBUTING.md
jadecarino: Its used to authenticate to GHCR and will then be used as the credentials to push to your user's GitHub Packages so I guess yes it needs to match your user so its allowed to push to your GHCR

jadecarino: I'll rewrite this to something like `This requires your GitHub username so you can authenticate to GitHub Container Registry and push images to your personal packages.`

CONTRIBUTING.md
jadecarino: done

CONTRIBUTING.md
```diff
@@ -0,0 +1,166 @@
...
+
+You will first need to create a new GitHub Personal access token (classic) or Fine-grained personal access token.
+
+To create a GitHub Personal access token (classic):
+1. Go to the Settings for your GitHub account > Developer settings > Personal access tokens > Tokens (classic). 
+1. Select Generate new token (classic).
+1. Select write:packages access and give the token a name.
+1. Copy the token as it will disappear, and add this to your repository secrets for WRITE_GITHUB_PACKAGES_TOKEN.
+
+<!-- To create a Fine-grained personal access token:
```
techcobweb, line 149: didn't know you could comment on markdown like this. Useful to know.

techcobweb (approved):

CONTRIBUTING.md
jadecarino: :)

PR #189, opened 2025-02-11
Add non-zOS IVTs that currently run on prod1 to the ivts module

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2135

## Changes
- Added the following IVTs to the ivts module:
  - ArtifactManagerIVT
  - HttpManagerIVT
  - DockerManagerIVT
- Added a custom buildSrc 'galasa.ivt' plugin containing the common gradle config like repositories, dependencies, and plugins that can be applied to IVTs to reduce duplicate config and remove the need to specify plugin versions everywhere
- Updated the set-version script to bump plugin versions in the buildSrc build.gradle instead of in IVT build.gradle files

No changes have been made to the IVTs themselves.


jadecarino (approved):
Great stuff! I really like the way you've restructured this project and it will make things a lot simpler as we keep adding tests in.

PR #193, opened 2025-02-14
Fixing Workflow warnings

jaydee029:
For issue galasa-dev/projectmanagement#2142

### DIFF
--- a/.github/workflows/buildutils.yaml
+++ b/.github/workflows/buildutils.yaml
@@ -6,6 +6,8 @@
 name: Buildutils Main Build
 
 on:
+  pull_request:
+    branches: [main]
   workflow_call:
 
 env:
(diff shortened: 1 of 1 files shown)


jaydee029:
Hey @jadecarino I looked into it and found out,
the warnings are due to the cache miss in go-setup step in the buildutils workflow.

I pointed it towards the `go.sum` file in each case as mentioned in the docs, however to test this the cache needs to be created at least by one successful workflow, then the subsequent workflows wont see that warning.

Other ways to fix the warnings 

1) set go cache to false
2) surpress warnings (not recommended)

jadecarino (approved):
Good find - I suppose the setup-go action was expecting the go.sum to be in the repo root but thats not the case for this repository.

Approved, so it can run the first time, and then I will trigger a second run to check the warnings have disappeared.

PR #199, opened 2025-02-18
IMS TM Manager (#2148)

ideeley:
Initial drop of the new IMS TM Manager. This contains what should be the final product code (subject to review and/or defects), build updates, and the first batch of JUnits.

Code is heavily based on the CICS TM Manager  - I basically just copied everything and then modified it. In many cases, this was just doing change/replace on "CICS"->"IMS", "TS"->"TM", "Region"->"System" (matching case), but there was also a lot of deletions. And the system logon code is very different.

There are multiple commits in this PR, and it should be easier to review the code in the order they were committed.

Matching issue [here](https://github.com/galasa-dev/projectmanagement/issues/2148)


modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager.ivt/build.gradle
```diff
@@ -0,0 +1,29 @@
...
+
+// Note: These values are consumed by the parent build process
+// They indicate which packages of functionality this OSGi bundle should be delivered inside,
+// or referenced from.
+// The settings here are gathered together by the build process to create a release.yaml file 
+// which gathers-up all the packaging metadata about all the OSGi bundles in this component.
+ext.projectName=project.name
+ext.includeInOBR          = true
+ext.includeInMVP          = true
+ext.includeInBOM          = false
```
techcobweb, line 23: You probably do want to include it in the BOM

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/ImsTerminal.java
```diff
@@ -0,0 +1,46 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import dev.galasa.framework.spi.ValidAnnotatedFields;
+
+/**
+ * A zOS 3270 Terminal for use with an IMS TM System that has access to the default IMS screens
+ * 
+ * <p>
```
techcobweb: FYI: These <p> markers do nothing if you already left a blank line. A blank line in Javadoc is treated as a new paragraph.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/ImsTerminal.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.FIELD })
+@ImstmManagerField
+@ValidAnnotatedFields({ IImsTerminal.class })
+public @interface ImsTerminal {
+
+    /**
+     * The tag of the IMS system terminal is to be associated with
```
techcobweb: Documenting that the default is "PRIMARY" may help anyone reading the javadoc.
Same for the other properties.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/ImsTerminal.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    String imsTag() default "PRIMARY";
+    
+    /**
+     * The IMS TM Manager will automatically connect the terminal to the IMS TM System when ever it starts
```
techcobweb: "whenever" is one word.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/ImstmManagerField.java
```diff
@@ -0,0 +1,24 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Used to annotate annotations that are to be used for Test Class fields. To be
+ * populated by the Manager.
+ *
+ *  
+ *
```
techcobweb: Any reason you leave a ton of space in the first javadoc comment ?
I see that elsewhere also. Wondered why.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/ImstmManagerField.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+package dev.galasa.imstm;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Used to annotate annotations that are to be used for Test Class fields. To be
```
techcobweb: I don't understand what this means. Could you describe in more detail pls ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmDefaultLogonProvider.java
```diff
@@ -0,0 +1,133 @@
...
+            imsTerminal.waitForTextInField(initialText);
+        } catch (Exception e) {
+            throw new ImstmManagerException(
+                    "Unable to logon to IMS, initial screen does not contain '" + initialText + "'");
+        }
+    }
+
+    private void waitForSignonScreen(IImsTerminal imsTerminal) throws ImstmManagerException {
+        try {
+            imsTerminal.waitForTextInField("DFS3649A");
```
techcobweb: Is this always the same value ? Making it a constant may help explain what it's for rather than a literal.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmDefaultLogonProvider.java
```diff
@@ -0,0 +1,133 @@
...
+        } catch (Exception e) {
+            throw new ImstmManagerException("Unable to wait for the initial IMS screen, looking for 'DFS3649I'",
```
techcobweb: The code waits for "DFS3649A" and if not found says it's looking for 'DFS3649I' ? Use a constant and reference it twice ?

techcobweb: Surprised this isn't configurable too, but I know little about IMS

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmDefaultLogonProvider.java
```diff
@@ -0,0 +1,133 @@
...
+                    e);
+        }
+    }
+
+    private void waitForSignedOnText(IImsTerminal imsTerminal) throws ImstmManagerException {
+
+        String[] pass = { "DFS3650I" };
+        String[] fail = { "REJECTED" };  // Any signon failure will give us DFS3649A again, with REJECTED and a reason code.
```
techcobweb: Can it be a foreign language, non-english ? Is it translated ever ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmDefaultLogonProvider.java
```diff
@@ -0,0 +1,133 @@
...
+            imsTerminal.waitForTextInField("DFS3649A");
+        } catch (Exception e) {
+            throw new ImstmManagerException("Unable to wait for the initial IMS screen, looking for 'DFS3649I'",
+                    e);
+        }
+    }
+
+    private void waitForSignedOnText(IImsTerminal imsTerminal) throws ImstmManagerException {
+
+        String[] pass = { "DFS3650I" };
```
techcobweb: Again, constants would be good rather than literals.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmDefaultLogonProvider.java
```diff
@@ -0,0 +1,133 @@
...
+        String[] fail = { "REJECTED" };  // Any signon failure will give us DFS3649A again, with REJECTED and a reason code.
+                                         // However, we can't search for DFS3649A as it's already on the screen.
+
+        try {
+            imsTerminal.waitForTextInField(pass, fail);
+        } catch (Exception e) {
+            throw new ImstmManagerException("Unable to sign on, looking for '" + String.join("', '", pass) + "'",
```
techcobweb: Don't understand why you are using String.join here... pass is a 1-element array, you could use pass[0] instead ?

Use `MessageFormat.format("Unable to sign on, looking for ''{0}''",pass[0]);` instead ?

techcobweb: Does the error get logged ? Might be good to log the exception prior to throwing it, or get the exception constructor to log ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
```diff
@@ -0,0 +1,360 @@
...
+import dev.galasa.textscan.spi.ITextScannerManagerSpi;
+import dev.galasa.zos.spi.IZosManagerSpi;
+import dev.galasa.zos3270.TerminalInterruptedException;
+
+@Component(service = { IManager.class })
+public class ImstmManagerImpl extends AbstractManager implements IImstmManagerSpi {
+    protected static final String NAMESPACE = "imstm";
+
+    private static final Log logger = LogFactory.getLog(ImstmManagerImpl.class);
+    private boolean required;
```
techcobweb: Better to call booleans isRequired. More standard for Java.
Also, what is required ? Something in the name of the variable to say what is required would be helpful.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
```diff
@@ -0,0 +1,360 @@
...
+    private IZosManagerSpi zosManager;
+    private ITextScannerManagerSpi textScanner;
+
+    private final HashMap<String, IImsSystemProvisioned> provisionedImsSystems = new HashMap<>();
+
+    private final ArrayList<IImsSystemProvisioner> provisioners = new ArrayList<>();
+    private final ArrayList<ImsTerminalImpl> terminals = new ArrayList<>();
+    private final ArrayList<IImsSystemLogonProvider> logonProviders = new ArrayList<>();
+
+    private String provisionType;
```
techcobweb: What does this mean ? What are the possible values ? would an enum be more appropriate ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
```diff
@@ -0,0 +1,360 @@
...
+    
+    @Override
+    public void initialise(@NotNull IFramework framework, @NotNull List<IManager> allManagers,
+            @NotNull List<IManager> activeManagers, @NotNull GalasaTest galasaTest) throws ManagerException {
+        super.initialise(framework, allManagers, activeManagers, galasaTest);
+        // *** Check to see if any of our annotations are present in the test class
```
techcobweb: Why the '***' ? adds to clutter.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
```diff
@@ -0,0 +1,360 @@
...
+
+    @Override
+    public List<String> extraBundles(@NotNull IFramework framework) throws ManagerException {
+        try {
+            ImstmPropertiesSingleton.setCps(framework.getConfigurationPropertyService(NAMESPACE));
+        } catch (ConfigurationPropertyStoreException e) {
+            throw new ImstmManagerException("Unable to request framework services", e);
+        }
+
+        return ExtraBundles.get();
```
techcobweb, line 116: I personally dislike the use of static methods, would rather it were injected/constructed at the start. But I know that's what you were copying... I realise many of my comments will be critisizing the CICS manager indirectly.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
```diff
@@ -0,0 +1,360 @@
...
+
+    @Override
+    public void provisionStop() {
+        for (ImsTerminalImpl terminal : this.terminals) {
+            try {
+                terminal.writeRasOutput();
+            	terminal.flushTerminalCache();
+                terminal.disconnect();
+            } catch (TerminalInterruptedException e) { // NOSONAR - wish to hide disconnect errors
+            }
```
techcobweb, line 284: Would be good to log disconnect errors ? Do they happen often ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
```diff
@@ -0,0 +1,360 @@
...
+    public void imstmSystemStarted(IImsSystem system) throws ImstmManagerException {
+        // A system has started, so connect everything up
+        
+        // Connect terminals that are associated with the system
+        
+        for(ImsTerminalImpl terminal : terminals) {
+            if (terminal.getImsSystem() == system) {
+                if (terminal.isConnectAtStartup()) {
+                    if (!terminal.connectToImsSystem()) {
+                        throw new ImstmManagerException("Failed to connect terminal to IMS TM system");
```
techcobweb: Won't `connectToImsSystem()` throw an exception if it fails to connect ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/IImsTerminal.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+package dev.galasa.imstm;
+
+import dev.galasa.zos3270.ITerminal;
+
+public interface IImsTerminal extends ITerminal {
+
+    IImsSystem getImsSystem();
+
+    boolean connectToImsSystem() throws ImstmManagerException;
```
techcobweb: If you tell it to connect, and it fails, why not throw an exception rather than return a boolean ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/dse/DseImsImpl.java
```diff
@@ -0,0 +1,58 @@
+/*
```
techcobweb, line 1: This class doesn't seem to actually do anything ? Why do we need it ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/dse/DseProvisioningImpl.java
```diff
@@ -0,0 +1,99 @@
...
+    private final ImstmManagerImpl imstmManager;
+
+    private final boolean enabled;
+
+    public DseProvisioningImpl(ImstmManagerImpl imstmManager) {
+        this.imstmManager = imstmManager;
+
+        String provisionType = this.imstmManager.getProvisionType();
+        switch (provisionType) {
+            case "DSE":
```
techcobweb, line 39: Ahh. ref previous comment. An enum would be better... ? Even if it just listed the types you supported. You could ask the enum if the provision type is found in the enum based on the name perhaps.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/dse/DseProvisioningImpl.java
```diff
@@ -0,0 +1,99 @@
...
+import dev.galasa.imstm.spi.IImsSystemProvisioner;
+import dev.galasa.zos.IZosImage;
+import dev.galasa.zos.ZosManagerException;
+
+public class DseProvisioningImpl implements IImsSystemProvisioner {
+    private static final Log logger = LogFactory.getLog(DseProvisioningImpl.class);
+
+    private final ImstmManagerImpl imstmManager;
+
+    private final boolean enabled;
```
techcobweb: In Java 'isEnabled' is better no ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/dse/DseProvisioningImpl.java
```diff
@@ -0,0 +1,99 @@
...
+        try {
+            zosImage = imstmManager.getZosManager().getImageForTag(imageTag);
+        } catch (ZosManagerException e) {
+            throw new ImstmManagerException("Unable to locate zOS Image tagged " + imageTag, e);
+        }
+
+
+        DseImsImpl imsSystem = new DseImsImpl(this.imstmManager, imsTag, zosImage, applid);
+
+        logger.info("Provisioned DSE " + imsSystem.toString() + " on zOS Image " + imsSystem.getZosImage().getImageID() + " for tag '" + imsSystem.getTag());
```
techcobweb: `MessageFormat.format`... is easier to read.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DefaultLogonInitialText.java
```diff
@@ -0,0 +1,21 @@
...
+
+import dev.galasa.imstm.ImstmManagerException;
+import dev.galasa.framework.spi.ConfigurationPropertyStoreException;
+import dev.galasa.framework.spi.cps.CpsProperties;
+
+public class DefaultLogonInitialText extends CpsProperties {
+
+    public static String get() throws ImstmManagerException {
+        try {
+            return getStringNulled(ImstmPropertiesSingleton.cps(), "default.logon", "initial.text");
```
techcobweb: Are all the IMS systems contacted going to have the same logon text ?
If one test connected to two systems and they had different logon text would it work ?

Singletons are generally an anti-pattern. Is there a better way of doing this ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DefaultVersion.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ */
+public class DefaultVersion extends CpsProperties {
+
+    private static final Log logger = LogFactory.getLog(DefaultVersion.class);
+
+    public static ProductVersion get() {
+        String version = "";
+        try {
+            version = getStringWithDefault(ImstmPropertiesSingleton.cps(), "15.5.0", "default", "version");
```
techcobweb: Better to set the version and use toString() on it rather than hard-code something in the message which may diverge ? Especially as this constant is regurgitated in multiple messages.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DefaultVersion.java
```diff
@@ -0,0 +1,52 @@
...
+            return ProductVersion.parse(version);
+        } catch (ImstmManagerException e) {
+            logger.error("Problem accessing the CPS for the default IMS version, defaulting to 15.5.0");
+            return ProductVersion.v(15).r(5).m(0);
```
techcobweb: Seems that a constant
```
private static final DEFAULT_PRODUCT_VERSION = new ProductVersion("15","5","0");
```
And then using that would be better to me.

Using ProductVersion.v(15).r(5).m(0); also creates 3 separate objects, discarding two of them... very inefficient.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DseApplid.java
```diff
@@ -0,0 +1,42 @@
...
+ * @galasa.examples 
+ * <code>imstm.dse.tag.PRIMARY.applid=IMS1</code><br>
+ *
+ */
+public class DseApplid extends CpsProperties {
+
+    public static String get(String tag) throws ImstmManagerException {
+        try {
+            String applid = getStringNulled(ImstmPropertiesSingleton.cps(), "dse.tag", "applid", tag);
+            return applid != null? applid.toUpperCase(): applid;
```
techcobweb: You might want to `.trim()` also to avoid any rogue whitespace at the start or end of the string ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DseVersion.java
```diff
@@ -0,0 +1,53 @@
...
+            if (version == null) {
+                return null;
+            } else {
+                return ProductVersion.parse(version);
+            }
+        } catch (ImstmManagerException e) {
+            // Prevent ImstmManagerExceptions from being caught by the more generic ManagerException
+            throw e;
+        } catch (ManagerException e) {
+            throw new ImstmManagerException("Failed to parse the IMS version '" + version + "' for tag '" + tag + "', format should be 0.0.0", e);
```
techcobweb: Instea of "0.0.0" , this might be better: "{integer}.{integer}.{integer} for example 15.5.0".

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ExtraBundles.java
```diff
@@ -0,0 +1,51 @@
...
+ * @galasa.valid_values bundle symbolic names comma separated
+ * 
+ * @galasa.examples <code>imstm.extra.bundles=org.example.imstm.provisioning</code><br>
+ *
+ */
+public class ExtraBundles extends CpsProperties {
+
+    public static List<String> get() throws ImstmManagerException {
+        try {
+            List<String> list = getStringList(ImstmPropertiesSingleton.cps(), "extra", "bundles");
```
techcobweb: 'list' isn't very descriptive. as a variable name.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/ImsTerminal.java
ideeley: It's exactly the same in CicsTerminal! For cleanliness I shall remove the tags and let the blank lines do the formatting. 

Not sure what purpose the trailing blank lines have so removing those too.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
```diff
@@ -0,0 +1,52 @@
...
+    public void activate() {
+        setInstance(this);
+    }
+    
+    @Deactivate
+    public void deactivate() {
+        setInstance(null);
+    }
+    
+    public static IConfigurationPropertyStoreService cps() throws ImstmManagerException {
```
techcobweb: Java naming conversions are `getCPS` ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
```diff
@@ -0,0 +1,52 @@
...
+
+import dev.galasa.imstm.ImstmManagerException;
+import dev.galasa.framework.spi.IConfigurationPropertyStoreService;
+
+@Component(service=ImstmPropertiesSingleton.class, immediate=true)
+public class ImstmPropertiesSingleton {
+    
+    private static ImstmPropertiesSingleton singletonInstance;
+
+    private static void setInstance(ImstmPropertiesSingleton instance) {
```
techcobweb, line 20: I'm surprised you didn't put anything in to synchronize multiple threads activate() and deactivate() at the same time, or multiple activates occurring concurrently...etc.
Singletons usually have such protections.

I guess in this case we know the OSGi framework won't try to do anything like that though... so may not be needed.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+    
+    @Deactivate
+    public void deactivate() {
+        setInstance(null);
+    }
+    
+    public static IConfigurationPropertyStoreService cps() throws ImstmManagerException {
+        if (singletonInstance != null) {
+            return singletonInstance.cps;
```
techcobweb: I would rather you check for null and throw the exception, before dropping through to a return at the end of the function. Return in the middle of the function looks odd when all we have is an exception afterwards.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+        
+        throw new ImstmManagerException("Attempt to access manager CPS before it has been initialised");
+    }
+    
+    public static void setCps(IConfigurationPropertyStoreService cps) throws ImstmManagerException {
+        if (singletonInstance != null) {
+            singletonInstance.cps = cps;
```
techcobweb, line 46: Is there a timing window here, when deactivate() is called, and sets the instance to null, just as you de-reference it between the if and the use of the cps method ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
```diff
@@ -0,0 +1,52 @@
...
+            return;
+        }
+        
+        throw new ImstmManagerException("Attempt to set manager CPS before instance created");
```
techcobweb: Again, return statement at the end of the method is better ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/ImsTerminal.java
ideeley: Added lines `Default value is <b>PRIMARY</b>.`, `Default value is <b>true</b>.` and `Default value is an empty string.`

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/ImsTerminal.java
ideeley: Fixed. Again, source is CicsTerminal!

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/spi/IImsSystemProvisioned.java
```diff
@@ -0,0 +1,19 @@
...
+import dev.galasa.imstm.ImstmManagerException;
+import dev.galasa.imstm.IImsSystem;
+
+public interface IImsSystemProvisioned extends IImsSystem {
+
+    String getNextTerminalId();
+    
+    
+    void submitRuntimeJcl() throws ImstmManagerException;
+    boolean hasRegionStarted() throws ImstmManagerException;
```
techcobweb: hasRegionStarted --> isRegionStarted ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/ImstmManagerField.java
ideeley: Because that's how it is in the CICS Manager code.... Fixed that one.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/spi/IImsSystemProvisioned.java
```diff
@@ -0,0 +1,19 @@
...
+
+import dev.galasa.imstm.ImstmManagerException;
+import dev.galasa.imstm.IImsSystem;
+
+public interface IImsSystemProvisioned extends IImsSystem {
+
+    String getNextTerminalId();
+    
+    
+    void submitRuntimeJcl() throws ImstmManagerException;
```
techcobweb: Where does the JCL come from ? What JCL gets submitted to the runtime?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/ImstmManagerField.java
ideeley: Again, simply copied from CicstsManagerField.

The ImsTerminal and ImsSystem annotation classes themselves are annotated with this one. I'm not sure how to expand on this without getting incredibly verbose.

Maybe something like "Used to annotate the classes that are used to annotate IMS TM fields in user test classes. These fields will be populated by the IMS TM Manager during initialization."?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmDefaultLogonProvider.java
ideeley: As far as I'm aware, it is. Even when security is not enabled for an IMS System, the user still has to supply a user id (but password is ignored), and the challenge is the same.  You can have terminals with preset security much like in CICS, but our 3270 terminals won't have that as we don't know the netname in advance. I've not found anything in documentation to suggest that this is customizable but, if it is, we'd need to do the whole setting up a new CPS property for it, much like I had to do with the default VTAM logon string.

I've made a constant called SIGNON_CHALLENGE for it. I've also moved pass and fail from waitForSignedOnText to be be class-scoped constants, renamed as SIGNON_SUCCESSFUL and SIGNON_FAILED (along with the comment to explain why we look for REJECTED when it fails) for consistency..

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmDefaultLogonProvider.java
ideeley: Yeah, done as part of the previous review comment fix. I can't claim that I know that much about IMS either, but what becomes pretty obvious very early on is that it is nowhere near as customizable as CICS.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmDefaultLogonProvider.java
ideeley: That's... a disturbingly good question.

ideeley: I can't find any IMS system startup parameters that allow a user to specify a language (I've looked through the documentation and also checked our systems to see if there were any parameters that specified "E" or "EN" - nothing. General google searches on IMS with language, chinese, japanese, etc. don't find anything either.

In the absence of any better information, I'm going to have to stick with English only. Worst case scenario is that our logic would take longer to fail i.e. wait for timeout instead of immediately spotting the failure message. If anybody complains, they should be able to supply the information needed to fix the problem, which should be comparatively straightforward as we can search for multiple failure strings - we'd just search for "REJECTED" in all languages rather than having to set up a cps property.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmDefaultLogonProvider.java
ideeley: Yep, preempted in response to your earlier comment.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmDefaultLogonProvider.java
ideeley: Again, copied from CICS....

It's more maintainable this way - if we ever need to search for multiple strings (though this seems unlikely), then we just add them to the list and this logic will pick them up automatically

The exception isn't caught before it gets back to the framework code so the message would be visible in run.log as part of the stack trace.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
ideeley: Copied from CicstsManagerImpl....

Renaming to isManagerRequired - and explicitly initializing to false rather than letting it default.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
ideeley: If you're referring to provisionType, it holds the value extracted from the imstm.provision.type cps property. It's typically going to be dse (or DSE) but it could be anything else if/when somebody writes their own IMS provisioner. I'll add a comment.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
ideeley: Because CicstsManagerImpl did it first! Removed.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
ideeley: Not going to disagree with you there.... I also find myself wondering why each property requires its own class. Could probably roll getExtraBundles(), getDseApplid(String), etc. methods into ImstmPropertiesSingleton (and get rid of all of its static methods as well (except for an instance() method to get the singleton).

If there's time after this PR has been merged, I can probably go through and fix all of that before we release 0.40.0.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
ideeley: Again copied from CicstsManagerImpl.

This is a zOS 3270 thing, I have no idea how it happens or how likely it is. I would suggest, however, that if it's going to be logged, it should be the zOS 3270 Manager doing it. As far as we're concerned, we're not using the terminal any more so we simply don't care what happens to it.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
ideeley: Again copied from CicstsManagerImpl....

You're correct that the default logon provider will throw an exception if it fails to connect. I'm reading this as belt and braces in case someone writes their own logon provider that behaves differently.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/IImsTerminal.java
ideeley: Copied from ICicsTerminal.

They're both valid ways of achieving the same end result and I'm not aware that either approach is considered better than the other (except that throwing and catching an exception is typically more expensive, if you care about the performance). In this case we're going to throw an exception anyway....

I'm more inclined to keep it as it is for consistency with the CICS support. However, if we're going to change it, it needs to be done before 0.40.0 goes out.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/dse/DseImsImpl.java
ideeley: Copied from DseCicsImpl....

It implements a number of methods specified by IImsSystemProvisioned that aren't implemented by BaseImsImpl. Most of them throw exceptions if they're called because DSE doesn't support them - getVersion() does real work though. This is all part of keeping IImsSystemProvisioned abstracted in case someone wants to write their own provisioner - I can't see how else we would do it.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/dse/DseProvisioningImpl.java
ideeley: As mentioned in the previous comment, provisionType might gain some extra valid values if someone writes their own provisioner. If we referenced it in multiple places we could map it onto an enum just for this class - which would require logic similar to this anyway - but we don't, so it's not worth the effort.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/dse/DseProvisioningImpl.java
ideeley: Blaming it on CICS again, but yes. Changed.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/dse/DseProvisioningImpl.java
ideeley: Blaming CICS again, but agreed. And was wrong anyway - missing the closing single quote after the tag name! Changed.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DefaultLogonInitialText.java
ideeley: Again copied from the CICS support.

On closer inspection, looking at ImstmDefaultLogonProvider (and CicstsDefaultLogonProvider), I think this should be a zOS property (we look for this string before issuing the LOGON APPLID command). In theory, if you've got two IMS systems (or two CICS regions) on different LPARs, then they _could_ have different initial text.

With your agreement, I'll change this to be a zOS property (similar to the custom VTAM logon string) and change the IMS code to use that instead (so this class will be deleted). Then, separately, we should change the CICS code to use the new property if it doesn't find the CICS-specific logon initial text property, and then deprecate the CICS property for later deletion.

I also don't understand why it's "default" logon initial text - that would imply that it can be overridden at the individual CICS region or IMS system level, but there's no logic in place to support that. A custom logon provider _could_, I suppose, but why would they? Perhaps, when creating new zos properties, we could do a "default" string for general use, and have an override property at the zOS image level?

I don't know what you mean by "anti-pattern" - but this would all be moot anyway if I change to zOS properties.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DefaultVersion.java
ideeley: Created a ProductVersion DEFAULT_VERSION constant and using as suggested - but copied from CICS again!

While we're in this area: Any thoughts on what version we should select as the default? For now, I've gone with the latest in-service release on the grounds that it will be valid for longer.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DefaultVersion.java
ideeley: Yeah, would be nice - but that constructor's not visible (and it's ints rather than Strings anyway)

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DseApplid.java
ideeley: Done - but, again, copied from CICS!

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DseVersion.java
ideeley: I'm sorry, but "{integer}.{integer}.{integer}" is **hideous**! I'm more inclined to go with the Javadoc text and say "a valid V.R.M version format, for example 15.5.0".

I suspect that "0.0.0" was chosen (by CICS...) so that it would never need to be updated.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ExtraBundles.java
ideeley: I'm blaming CICS again - but changing to `bundles`.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
ideeley: Again, copied from CICS. Since this one is API that could be used by anyone writing their own IMS provisioner, I'm inclined to keep this one as it is for consistency with the CICS manager.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
ideeley: If CICS didn't do it, I haven't done it.

Even if there were multiple threads doing this, the end result would still be the same as if it were synchronized provided the last setCps call is made after the last activate. And it definitely doesn't matter if deactivates interleave. There could only be an issue if activates and deactivates are interleaving - but you've got chaos there even if they're synchronized.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
ideeley: Agreed but, y'know, copied from CICS and all that.... Changed. setCps too.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
ideeley: So we'd get an NPE instead of an ImstmManagerException. Frankly, I'd be more concerned about why OSGi is deactivating my service when I'm clearly still using it.

However, I've seen a straightforward fix so done that:
```
        try {
            singletonInstance.cps = cps;
        } catch (NullPointerException e) {
            throw new ImstmManagerException("Attempt to set manager CPS before instance created");
        }
```

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
ideeley: Changed for earlier review comments.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/spi/IImsSystemProvisioned.java
ideeley: That would come from a custom provisioner.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/ImstmManagerField.java
techcobweb: Clearer on re-reading. Ignore or add your own (better) explanation. Just wasn't sure what was annotating what.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DefaultLogonInitialText.java
techcobweb: FYI: anti-pattern w.r.t Singleton... so I try not to create them.

https://medium.com/aia-sg-techblog/why-singleton-pattern-is-considered-as-anti-design-pattern-c81dd8b7e757

There are many similar views out there on Singletons.

I agree with your 'default' argument. Happy if it changes. And your proposal to move the CPS property (so long as it doesn't break people immediately).

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DefaultVersion.java
techcobweb: I see. Someone obviously got a bit excited about the fluent call pattern. I think I'll raise a defect and fix the code so it doesn't new up 2 extra objects, and you can leave this code as-is. Done. See ProductVersion constructors are very inefficient.
[#2151](https://github.com/galasa-dev/projectmanagement/issues/2151)

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DseVersion.java
techcobweb: > is hideous! 

Not sure I agree, as it's pretty clear that 3 integers are needed, but your suggestion is way better than 0.0.0.
Might be that it never needs updating, but at least it might resonate with people as a valid version. Go with your suggestion ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/test/java/dev/galasa/tests/TestGoldenPath.java
```diff
@@ -0,0 +1,97 @@
...
+import dev.galasa.framework.spi.IManager;
+import dev.galasa.framework.spi.language.GalasaTest;
+import dev.galasa.testharness.TestHarnessFramework;
+import dev.galasa.zos.IZosImage;
+import dev.galasa.zos.spi.IZosManagerSpi;
+import dev.galasa.zosbatch.spi.IZosBatchSpi;
+import dev.galasa.zosfile.spi.IZosFileSpi;
+
+//@RunWith(PowerMockRunner.class)
+public class TestGoldenPath {
```
techcobweb, line 31: Why all commented-out ?
Not much point having it here if it's not testing anything.

ideeley: Copied from CICS. I'm guessing commented out because Powermock isn't available. I created the commented out class based on the CICS original for if/when Powermock is made available.

However, since most of the capability of Powermock is now available to us in the uplifted Mockito, I can revisit this - but it's lower down the list of priorities.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/ImstmManagerField.java
ideeley: Updated with my "Maybe..." suggestion.

techcobweb:
Whew. Got to the end. That's my 1st-pass through. 

Great to see this new code being developed.

I see you've started changing/fixing also, will take another look early next week to see what's changed...

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/spi/IImsSystemProvisioned.java
ideeley: Copied from CICS....

And is an instance where I missed my global change/replace. Changed to isSystemStarted().

techcobweb:
FYI. If you pull the latest, the ProductVersion(int,int,int) constructor is visible now.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DefaultLogonInitialText.java
ideeley: Property changed to be a zOS Manager property (it's still using static methods and a singleton because that's just how the zOS Manager has been implemented and there's no way I'm changing that).

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/DseVersion.java
ideeley: Done

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
techcobweb: OK. That sounds like a reasonable trade-off.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
techcobweb: My point here was that the exception was being swallowed rather than thrown, so the calling code can't log it...

techcobweb: You will care if resources are not freed when this happens. Trying to debug something when there is no mention of a failure in the code is harder.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
techcobweb: Seems that then there are 2 patterns for errors. Throwing, or returning false. I would scratch the last one as you don't get as much diagnostics information out.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/IImsTerminal.java
techcobweb: I do care about performance, but in error cases performance doesn't matter as much as being able to find errors and fix them quickly.

I think changing it prior to 0.40.0 should be done.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/dse/DseImsImpl.java
techcobweb: Not sure I understand still. Lets talk.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
techcobweb: Noooo. I'd be more inclined to do it properly, and raise an issue to fix up the CICS one, and deprecate this method with the bad names. Changing now will mean the IMS one doesn't need to do all that ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
techcobweb: There is no way any deactivates would interleave, as only the last one to set itself will be available.

This is one more problem that singletons have. Instance is a superior pattern.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
techcobweb: Thanks.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/spi/IImsSystemProvisioned.java
techcobweb: How do they inject Jcl into the code then ?

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/test/java/dev/galasa/tests/TestGoldenPath.java
techcobweb: OK. Thanks

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
techcobweb: As per our discussion. Lets leave this as-is for now.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
ideeley: You're talking about logging at a higher level. I'm saying that it should be logged at a lower level - _before_ the exception is thrown.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/spi/IImsSystemProvisioned.java
ideeley: Not sure I understand the question. Without fully designing a custom provisioner, I think you'd be looking at a submitRuntimeJcl() method that does something like:

- extract a property that points to a JCL template file
- read in the template
- substitute values into the template specific to this system e.g. applid, jobname, credentials
- submit the job using the zOS Batch Manager

I don't know what else you're looking for.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/IImsTerminal.java
ideeley: OK, method signature changed to void rather than boolean, and the exception thrown in ImstmManagerImpl.imstmSystemStarted() on receiving false as a response has been moved into ImsTerminalImpl.connectToImsSystem() instead of returning false

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/properties/ImstmPropertiesSingleton.java
ideeley: OK, changed - subject to a later complete rewrite....

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/test/java/dev/galasa/tests/TestGoldenPath.java
ideeley: As mentioned in person earlier, this test wasn't using Powermock at all! Didn't take an awful lot to clean it up and get it working.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
techcobweb: changed to void.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/spi/IImsSystemProvisioned.java
techcobweb: Not needed.

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmManagerImpl.java
ideeley: Warning now logged in Terminal.disconnect() - so CICS Manager will inherit this too.

techcobweb (approved):
All looks good to me.

PR #200, opened 2025-02-18
more flexible way of initialising the framework and customizing what gets set and what doesnt based on use it will be put to.

techcobweb:

# Why ?

Part of the changes related to this story: [Ability to access managers during Resource Management #2124](https://github.com/galasa-dev/projectmanagement/issues/2124)

- Needed a more flexible way of initialising the framework without letting the framework 
know what type of framework it is.

- Added Init strategy interface and 3 implementations of it, called from different places depending on which server is being created.
- Added a Factory class so you can get at the implementations of the strategies without having direct access to them
- Intention is to move as much as we can to internal and provide access via interfaces and this GalasaFactory class

Intended as an alternative to https://github.com/galasa-dev/galasa/pull/180


eamansour (approved):
Can't spot anything immediately wrong. Nice separation of logic out from the initialisation class - approved

PR #206, opened 2025-02-20
ProductVersion had a few deficiencies

techcobweb:

# Why ?
Goes to address [ProductVersion constructors are very inefficient. #2151](https://github.com/galasa-dev/projectmanagement/issues/2151)

- Efficient constructor made public. Why wasn't it in the first place ? No idea.
- Checks for negative numbers (had to coerce them to zero if using the v() r() or m() methods because I couldn't throw the exception as that would change the signature and make it backwards incompatible). It's a bit crappy but it means -ve numbers aren't useable with eithe r construction method.
- Recommend using the public method with all 3 parameters, more efficient that way.
- Shied away from deprecating the old way of doing it. Though I could be persuaded.
- Oh, and it was missing a hashCode() method which should always have been there, as equals() was implemented ! Really poor rookie mistake there.
- Added some unit tests. Again. Rookie mistake.

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa/src/main/java/dev/galasa/ProductVersion.java
+++ b/modules/framework/galasa-parent/dev.galasa/src/main/java/dev/galasa/ProductVersion.java
@@ -5,11 +5,20 @@
  */
 package dev.galasa;
 
+import java.text.MessageFormat;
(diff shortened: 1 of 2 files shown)


jadecarino (approved):
Nice, approved

PR #208, opened 2025-02-26
Remove hard coded value in Docker Manager for busybox image name

jadecarino:
## Why?

The Docker Manager has a hard coded string for the fully qualified name of a busybox Docker image at `library/busybox:latest`. This should not be hard coded and should be configurable as it only fits the naming convention for Dockerhub and might do other Docker registries. However, if a user of the Docker Manager has a 'busybox' image in their Docker registry in any other org or with another tag, the image won't be found and the test using the Docker Manager will fail.

This is the behaviour we have seen in our DockerLocalJava11Ubuntu test.

### DIFF
--- a/modules/managers/galasa-managers-parent/galasa-managers-cloud-parent/dev.galasa.docker.manager/src/main/java/dev/galasa/docker/internal/properties/DockerRegistryBusyboxImage.java
+++ b/modules/managers/galasa-managers-parent/galasa-managers-cloud-parent/dev.galasa.docker.manager/src/main/java/dev/galasa/docker/internal/properties/DockerRegistryBusyboxImage.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package dev.galasa.docker.internal.properties;
+
+import dev.galasa.docker.DockerManagerException;
+import dev.galasa.framework.spi.ConfigurationPropertyStoreException;
+import dev.galasa.framework.spi.cps.CpsProperties;
+
+/**
+ * Docker Registry Busybox Image CPS property
+ * 
+ * @galasa.cps.property
+ * 
+ * @galasa.name docker.registry.[registryId].busybox.image
+ * 
+ * @galasa.description Provides fully qualified image name including repository and tag of a 'busybox' Docker image on a given registry.
+ * 
+ * @galasa.required No - if not provided it will default to 'library/busybox:latest' which is compatible with the default Docker registry, Dockerhub.
+ * 
+ * @galasa.default library/busybox:latest
+ * 
+ * @galasa.valid_values library/busybox:latest, myorg/busybox:1.1.1
+ * 
+ * @galasa.examples 
+ * <code>docker.registry.[registryId].busybox.image=registryorg/busybox:latest</code>
+ * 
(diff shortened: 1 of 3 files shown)


modules/managers/galasa-managers-parent/galasa-managers-cloud-parent/dev.galasa.docker.manager/src/main/java/dev/galasa/docker/internal/properties/DockerRegistryBusyboxImage.java
```diff
@@ -0,0 +1,58 @@
...
+ * may not be in the 'library' namespace and may not be tagged with 'latest'.<br> 
+ * */
+public class DockerRegistryBusyboxImage extends CpsProperties {
+
+    public static String get(String[] dockerRegistries) throws DockerManagerException {
+		String busyboxImage = "";
+		if (dockerRegistries.length > 0) {
+			for (String dockerRegistry : dockerRegistries) {
+				try {
+					busyboxImage = getStringNulled(DockerPropertiesSingleton.cps(), "registry", "busybox.image", dockerRegistry);
```
eamansour: Really minor but instead of calling DockerPropertiesSingleton.cps() each time we loop, we could store a reference to it outside the loop and use that

modules/managers/galasa-managers-parent/galasa-managers-cloud-parent/dev.galasa.docker.manager/src/main/java/dev/galasa/docker/internal/properties/DockerRegistryBusyboxImage.java
```diff
@@ -0,0 +1,58 @@
...
+					throw new DockerManagerException("Problem asking the CPS for the Busybox image name on registry ID: "  + dockerRegistry, e);
+				}
+				if (busyboxImage != null && !busyboxImage.equals("")) {
+					break;
+				}
+			}
+		}
+		if (busyboxImage == null || busyboxImage.equals("")) {
+			busyboxImage = "library/busybox:latest";
+		}
```
eamansour: If the busyboxImage value is a string of spaces, it won't be set this default value - you can either do a `.trim().equals("")` or a `.isBlank()` to sort this out :)

eamansour (approved):
A couple of minor comments but approved assuming they'll be addressed

modules/managers/galasa-managers-parent/galasa-managers-cloud-parent/dev.galasa.docker.manager/src/main/java/dev/galasa/docker/internal/properties/DockerRegistryBusyboxImage.java
jadecarino: Good point - done

modules/managers/galasa-managers-parent/galasa-managers-cloud-parent/dev.galasa.docker.manager/src/main/java/dev/galasa/docker/internal/properties/DockerRegistryBusyboxImage.java
jadecarino: Thanks great point, done

PR #213, opened 2025-02-28
Refactor of IMS TM properties + minor change to CICS logon property usage

ideeley:
This is the refactor of the IMS TM properties. No more singleton, no more static methods, just a properties POJO that's passed around as needed.

Additionally, I've removed the extra bundles and default version properties in their entirety. The DSE provisioner doesn't use them - if a custom provisioner wants them, they can create their own property.

Bonus commit: Changing the CICS default logon provider to use the z/OS logon initial text property if - and only if - the CICS default logon initial text property is not set.


ideeley:
Forgot to reference the story. Here it is - https://github.com/galasa-dev/projectmanagement/issues/2148

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmProperties.java
```diff
@@ -0,0 +1,139 @@
...
+     * @galasa.valid_values A valid VTAM applid
+     * 
+     * @galasa.examples 
+     * <code>imstm.dse.tag.PRIMARY.applid=IMS1</code><br>
+     *
+     */
+    public String getDseApplid(String tag) throws ImstmManagerException {
+        if (!isSetDseApplid) {
+            try {
+                dseApplid = getStringNulled(cps, "dse.tag." + tag, "applid");
```
jadecarino, line 56: For consistency, I believe this line should match the use of `getStringNulled` on line 92, so something like: `getStringNulled(cps, "dse.tag", "applid", tag);`. Rather than appending the tag to `dse.tag`, use the suffixes parameter of `getStringNulled`.

jadecarino: Appreciate that the CICS TS Manager does it this way but that should probably be changed for consistency also - the majority of other places that call these CpsProperties methods, they don't tend to append strings together with `+`

jadecarino (changes requested):
One minor change for consistency in the new properties class, but otherwise looks good. Happy to approve after that change. Thanks!

modules/managers/galasa-managers-parent/galasa-managers-imstm-parent/dev.galasa.imstm.manager/src/main/java/dev/galasa/imstm/internal/ImstmProperties.java
ideeley: That's something that I very deliberately changed....
Coded how you have it, it will accept imstm.dse.tag.[TAG].applid, but it will also accept imstm.dse.tag.applid, which makes no sense. It's valid for version, because the concept of a default version works, but not for an applid.

jadecarino: Discussed the reason for this with Ian - fine with keeping it this way.

jadecarino (approved):

PR #230, opened 2025-03-11
build-locally scripts can now build the galasa-boot-embedded image locally

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2161

## Changes
- Added an optional `--docker` flag to the build-locally script in tools and the OBR module that results in the galasa-boot-embedded image being built locally
- Added a `setup-minikube-docker-registry.sh` script to perform the relevant setup required for minikube to be able to pull locally-built images when running a local Galasa service

### DIFF
--- a/tools/setup-minikube-docker-registry.sh
+++ b/tools/setup-minikube-docker-registry.sh
@@ -253,3 +253,4 @@ fi
 push_galasa_images_to_local_registry
 
 success "Local Docker registry has been successfully set up and populated with locally-built Galasa images!"
+info "If you wish to install a local Galasa service on minikube using the Helm chart, set the 'galasaRegistry' Helm value to 'localhost:5000' so minikube can pull the images from the local registry."
--- a/tools/setup-minikube-docker-registry.sh
+++ b/tools/setup-minikube-docker-registry.sh
@@ -134,6 +134,9 @@ function check_tool_is_installed {
 #-----------------------------------------------------------------------------------------
 function check_docker_installed {
     check_tool_is_installed "docker"
+    docker version > /dev/null 2>&1
+    rc=$?
+    check_exit_code ${rc} "The docker daemon is not running. Start it and try again."
 }
 
 #-----------------------------------------------------------------------------------------
--- a/tools/setup-minikube-docker-registry.sh
+++ b/tools/setup-minikube-docker-registry.sh
@@ -11,6 +11,8 @@
 # Objectives: Sets up and populates a local Docker registry with Docker images used for
 # Galasa development.
 #
+# Note: This script has only been tested on macOS.
+#
 #-----------------------------------------------------------------------------------------
 
 # Where is this script executing from ?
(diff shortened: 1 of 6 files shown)


tools/setup-minikube-docker-registry.sh
```diff
@@ -0,0 +1,256 @@
...
+
+#-----------------------------------------------------------------------------------------
+# Functions
+#-----------------------------------------------------------------------------------------
+function clean_up_before_set_up {
+    h2 "Cleaning up any existing Docker containers..."
+
+    docker stop docker-to-minikube-socat | xargs docker rm
+
+    success "Clean-up completed OK"
```
jadecarino, line 94: When I ran this script myself I got the below:
```
Cleaning up any existing Docker containers...
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
[✔] Clean-up completed OK
```
Maybe there should be an extra check before trying to use `docker` to check if the docker daemon is running?

tools/setup-minikube-docker-registry.sh
```diff
@@ -0,0 +1,256 @@
...
+    rc=$?
+    check_exit_code ${rc} "Failed to push the ${retagged_image_name} image to the local Docker registry"
+    success "Image ${retagged_image_name} pushed to local Docker registry OK"
+}
+
+#-----------------------------------------------------------------------------------------
+function setup_local_registry_for_minikube {
+    h2 "Configuring minikube to be able to pull images from a local Docker registry..."
+    minikube_ip=$(minikube ip)
+    rc=$?
```
jadecarino, line 210: Similarly I got the same error about docker daemon not running here:
```
Configuring minikube to be able to pull images from a local Docker registry...
❌  Exiting due to GUEST_STATUS: Unable to get control-plane node minikube host status: state: unknown state "minikube": docker container inspect minikube --format={{.State.Status}}: exit status 1
stdout:


stderr:
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

jadecarino (changes requested):
This looks great! Two comments (although these would be fixed from the same change) based on when I ran it myself - checking for the dameon running first might be a good addition.

As discussed I also got an error as I didn't have your changes to the webui script initially but I don't think thats a big issue, so no need to add anything for that one.

tools/setup-minikube-docker-registry.sh
eamansour: Thanks, added a check at the start of the script to check that the docker daemon is running in https://github.com/galasa-dev/galasa/pull/230/commits/cf29af5153ba3a84eded76c18329264971891f90

tools/setup-minikube-docker-registry.sh
eamansour: Extra docker daemon check added in https://github.com/galasa-dev/galasa/pull/230/commits/cf29af5153ba3a84eded76c18329264971891f90

jadecarino (approved):
Looks good to me!

PR #236, opened 2025-03-26
Add the ability to load custom resource management providers

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2067

## Changes
- Added functionality to pull a test stream, includes, and excludes from environment variables
- Resource management now loads bundles with the IResourceManagementProvider service from a given test stream as well as ones in the uber OBR, and the providers are filtered based on the provided includes and excludes regex lists


modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/MonitorConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+    private List<Pattern> convertListToPatternList(List<String> regexList) throws FrameworkException {
+        List<Pattern> patternList = new ArrayList<>();
+        for (String pattern : regexList) {
+            String patternToAdd = pattern;
+
+            // Patterns like '*' and '*mysuffix' are not valid regex patterns alone, so convert them into valid
+            // patterns like '.*' and '.*mysuffix' to accept any character (.) zero or more times (*) in prefixes.
+            if (pattern.startsWith("*")) {
+                patternToAdd = "." + pattern;
+            }
```
eamansour: Is this necessary? If we say we only expect valid regex patterns, then this won't be needed since a pattern like `*myMonitorClass` isn't valid.

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/MonitorConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public List<Pattern> getIncludesRegexList() {
+        return includesRegexList;
+    }
+
+    public List<Pattern> getExcludesRegexList() {
+        return excludesRegexList;
+    }
+
+    private List<Pattern> convertListToPatternList(List<String> regexList) throws FrameworkException {
```
eamansour: Should this pattern compilation be moved somewhere else?

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/MonitorConfiguration.java
techcobweb: Think it depends if we want to support glob matching or regex. Seems that glob matches should be sufficient so supporting *.x.y.z should probably match .*[.]x[.]y[.]z ?

techcobweb: glob matches are easier for people to get right

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/MonitorConfiguration.java
techcobweb: Possibly, some class that is passed the includes/excludes, then you can ask it isThisClassAllowed(xxx) or something ? Then we can test it in isolation.

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/MonitorConfiguration.java
techcobweb: BUT. It's a bit odd only processing the first 'glob' pattern and leaving the others alone. And not converting '.' in the input to '[.]' in the regex.
Why only do this for the startsWith() string, rather than all the way through ?

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/MonitorConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+            // Patterns like '*' and '*mysuffix' are not valid regex patterns alone, so convert them into valid
+            // patterns like '.*' and '.*mysuffix' to accept any character (.) zero or more times (*) in prefixes.
+            if (pattern.startsWith("*")) {
+                patternToAdd = "." + pattern;
+            }
+
+            try {
+                patternList.add(Pattern.compile(patternToAdd));
+            } catch (PatternSyntaxException e) {
+                throw new FrameworkException("Invalid regex pattern provided", e);
```
techcobweb: Think we should use glob matching.

techcobweb: Also, because if the user supplies a weird regex using lots of constructs could it crash or slow things down ? Perhaps we only allow * wildcard and document it.

I'm thinking the precedent is the OSGi bundle includes/excludes thing in terms of which classes are visible outside the bundle.

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/ResourceManagement.java
```diff
@@ -161,6 +196,102 @@ public void run(Properties bootstrapProperties, Properties overrideProperties) t
...
+    private void loadRepositoriesFromStream(String streamName, IStreamsService streamsService) throws FrameworkException {
+        IStream stream = streamsService.getStreamByName(streamName);
+
+        // Add the stream's OBR to the repository admin
+        String commaSeparatedObrs = stream.getObrLocation();
+        if (commaSeparatedObrs == null || commaSeparatedObrs.isBlank()) {
+            throw new FrameworkException("No OBR has been configured into the provided test stream");
+        }
+
+        for (String obr : commaSeparatedObrs.split(",")) {
```
techcobweb: I would prefer to see a method which gets an array of OBRs, separate from a loop which adds those to the repository admin.

Comma-separator list handling seems noise in this method when it could be moved to  a dedicated method devoted to get the list of OBRs from the stream.

Why can't the stream object IStream return a list ? seems strange that the function isn't pushed into there.

Also getObrLocation() is a weird name if it can return a comma separated list. Re-name to something more appropriate, or do as suggested above ?

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/ResourceManagement.java
```diff
@@ -161,6 +196,102 @@ public void run(Properties bootstrapProperties, Properties overrideProperties) t
...
+            try {
+                repositoryAdmin.addRepository(obr);
+            } catch (Exception e) {
+                throw new FrameworkException("Unable to load repository " + obr, e);
+            }
+        }
+
+        // Add the stream's maven repo to the maven repositories
+        String mavenRepo = stream.getMavenRepositoryUrl();
```
techcobweb: Seems this could be put into a stream.isValid() method ?

techcobweb: And we could have stream.getMavenRepositoryUrl(); return a URL, with a separate stream.getMavenRepositoryUrlAsString(); then if it wasn't a URL it would blow up before we try to use it, without this code worrying about the MalformedURLException

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/ResourceManagement.java
```diff
@@ -161,6 +196,102 @@ public void run(Properties bootstrapProperties, Properties overrideProperties) t
...
+    }
+
+    private Set<String> getResourceMonitorsFromRepository(Repository repository) {
+        Set<String> resourceMonitorBundles = new HashSet<>();
+        for (Resource resource : repository.getResources()) {
+            if (resource.getCapabilities() != null) {
+                for (Capability capability : resource.getCapabilities()) {
+                    if (isResourceMonitorCapability(capability)) {
+                        resourceMonitorBundles.add(resource.getSymbolicName());
+                        break;
```
techcobweb: this break isn't clear if you intend breakign out of the inner or outer loop. I always have this issue with nested loops and breaks.
Solution is to put the inner loop into a separate method... like
bool isResourceContainingAResourceMonitor(resource) ?

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/runner/TestRunnerDataProvider.java
```diff
@@ -7,6 +7,7 @@
 
 import java.util.Properties;
 
+import dev.galasa.framework.BundleManager;
```
techcobweb, line 10: Bit odd adding a new import but no code which uses it ?

modules/framework/galasa-parent/dev.galasa.framework/src/testFixtures/java/dev/galasa/framework/mocks/MockIConfigurationPropertyStoreService.java
```diff
@@ -25,12 +33,20 @@ public class MockIConfigurationPropertyStoreService implements IConfigurationPro
     @Override
     public @NotNull Map<String, String> getPrefixedProperties(@NotNull String prefix)
```
techcobweb, line 34: I'm growing to like the @NotNull and @Null annotations. Useful.

modules/framework/galasa-parent/galasa-boot/src/main/java/dev/galasa/boot/Launcher.java
```diff
@@ -653,4 +655,27 @@ public void setStoresFromEnvironmentVariables(Environment env, Properties bootst
             bootstrap.setProperty("framework.auth.store",authStore);
         }
     }
+
+    private MonitorConfiguration getResourceMonitorConfigFromEnvironment(Environment env) {
```
techcobweb: I might be tempted to move this into the MonitorConfiguration itself, passing the env to it... to keep parsing as close to the code using it, and unit tests of the parsing together...

modules/framework/galasa-parent/galasa-boot/src/main/java/dev/galasa/boot/MonitorConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class MonitorConfiguration {
+    
+    private String stream;
+    private List<String> includesRegexList;
+    private List<String> excludesRegexList;
+
+    public MonitorConfiguration(String stream, List<String> includesRegexList, List<String> excludesRegexList) {
+        this.stream = stream;
+        this.includesRegexList = includesRegexList;
```
techcobweb: I guess we should call them globPatterns or something if we move away from regexes ?

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/runner/TestRunnerDataProvider.java
eamansour: The BundleManager class was moved, the import was just updated to reflect its new location.

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/MonitorConfiguration.java
eamansour: Thanks, switched to glob matching in https://github.com/galasa-dev/galasa/pull/236/commits/fa36d081355b9cd4e5ef366e28cde2f8089f0b53

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/MonitorConfiguration.java
eamansour: Moved into a MonitorFilter class with unit tests to accompany it in https://github.com/galasa-dev/galasa/pull/236/commits/fa36d081355b9cd4e5ef366e28cde2f8089f0b53

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/MonitorConfiguration.java
eamansour: Switched to glob matching in https://github.com/galasa-dev/galasa/pull/236/commits/fa36d081355b9cd4e5ef366e28cde2f8089f0b53

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/ResourceManagement.java
eamansour: Clarified in https://github.com/galasa-dev/galasa/pull/236/commits/fa36d081355b9cd4e5ef366e28cde2f8089f0b53

modules/framework/galasa-parent/galasa-boot/src/main/java/dev/galasa/boot/Launcher.java
eamansour: Moved in https://github.com/galasa-dev/galasa/pull/236/commits/fa36d081355b9cd4e5ef366e28cde2f8089f0b53

modules/framework/galasa-parent/galasa-boot/src/main/java/dev/galasa/boot/MonitorConfiguration.java
eamansour: Done in https://github.com/galasa-dev/galasa/pull/236/commits/fa36d081355b9cd4e5ef366e28cde2f8089f0b53

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/ResourceManagement.java
eamansour: Added isValid() method and changed getMavenRepositoryUrl() return type in https://github.com/galasa-dev/galasa/pull/236/commits/d7bf0fdb58a20eeca9b4d071476e26ee6797180d

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/ResourceManagement.java
eamansour: Sounds good - updated the IStream object to return a List of OBRs and its setter handles the splitting of OBRs in https://github.com/galasa-dev/galasa/pull/236/commits/d7bf0fdb58a20eeca9b4d071476e26ee6797180d

modules/framework/galasa-parent/dev.galasa.framework.api.streams/src/main/java/dev/galasa/framework/api/streams/internal/routes/StreamsTransform.java
```diff
@@ -76,19 +77,19 @@ private StreamData createStreamData(IStream stream) {
 
         StreamData data = new StreamData();
         List<StreamOBRData> streamObrData = new ArrayList<>();
-        String testCatalogUrl = stream.getTestCatalogUrl();
-        String obrLocation = stream.getObrLocation();
+        URL testCatalogUrl = stream.getTestCatalogUrl();
```
techcobweb, line 79: What if the list of OBRs or the single OBR has a trailing slash ?

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/MonitorFilter.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+import dev.galasa.framework.spi.FrameworkException;
+
+public class MonitorFilter {
```
techcobweb: This is a classNameFilter ?

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/ResourceManagementProviders.java
```diff
@@ -85,37 +83,18 @@ private Set<IResourceManagementProvider> getResourceManagementProviders(
...
-        for (IResourceManagementProvider provider : providersToInclude) {
-            for (Pattern excludePattern : monitorConfig.getExcludesRegexList()) {
-                String providerName = provider.getClass().getCanonicalName();
-                Matcher excludeMatcher = excludePattern.matcher(providerName);
-                if (excludeMatcher.matches()) {
-                    providersToExclude.add(provider);
-                    break;
-                }
+            String monitorClassName = provider.getClass().getCanonicalName();
+            if (filter.isMonitorClassAllowed(monitorClassName)) {
```
techcobweb: Method could be less specific. isPassesConstraints() ? Have a think and come up with something better ?

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/ResourceManagement.java
```diff
@@ -219,13 +218,14 @@ void loadMonitorBundles(IBundleManager bundleManager, String stream, IStreamsSer
     private void loadRepositoriesFromStream(String streamName, IStreamsService streamsService) throws FrameworkException {
         IStream stream = streamsService.getStreamByName(streamName);
 
-        // Add the stream's OBR to the repository admin
-        String commaSeparatedObrs = stream.getObrLocation();
-        if (commaSeparatedObrs == null || commaSeparatedObrs.isBlank()) {
-            throw new FrameworkException("No OBR has been configured into the provided test stream");
+        if (!stream.isValid()) {
```
techcobweb: stream.validate() instead ? Then the exception can be way more specific.

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/MonitorFilter.java
```diff
@@ -0,0 +1,85 @@
...
+                Pattern patternToAdd = Pattern.compile(convertedPattern);
+                patternList.add(patternToAdd);
+            } catch (PatternSyntaxException e) {
+                throw new FrameworkException("Failed to compile glob pattern into a valid regex pattern", e);
+            }
+        }
+        return patternList;
+    }
+
+    private String getGlobAsRegexString(String globPattern) {
```
techcobweb: JavaClassGlobPattern class needs to process a single glob, turn to regex, fail to construct if it's invalid...etc.

techcobweb: only supporting * and ? excluding any characters which are not fully qualified java class names

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/streams/Stream.java
```diff
@@ -32,22 +41,33 @@ public void setDescription(String description) {
...
+    public void setTestCatalogUrl(String testCatalogUrl) throws StreamsException {
+        try {
+            this.testCatalogUrl = new URL(testCatalogUrl);
+        } catch (MalformedURLException e) {
+            throw new StreamsException("Invalid testcatalog URL provided", e);
+        }
     }
 
+    @Override
     public boolean getIsEnabled() {
```
techcobweb: isEnabled ??

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/streams/Stream.java
```diff
@@ -56,12 +76,27 @@ public void setIsEnabled(boolean isEnabled) {
...
+        if (commaSeparatedObrs != null && !commaSeparatedObrs.isBlank()) {
+            obrs = Arrays.asList(commaSeparatedObrs.split(","));
+        }
+        this.obrs = obrs;
     }
 
-    public void setObrLocation(String obrLocation) {
-        this.obrLocation = obrLocation;
+    @Override
+    public boolean isValid() {
```
techcobweb: will change to validate() method

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/streams/Stream.java
```diff
@@ -56,12 +76,27 @@ public void setIsEnabled(boolean isEnabled) {
...
-        return this.obrLocation;
+    @Override
+    public List<String> getObrs() {
+        return this.obrs;
+    }
+
+    public void setObrs(String commaSeparatedObrs) {
+        List<String> obrs = new ArrayList<>();
+        if (commaSeparatedObrs != null && !commaSeparatedObrs.isBlank()) {
+            obrs = Arrays.asList(commaSeparatedObrs.split(","));
```
techcobweb: needs to be trimmed
can we remove trailing slashes ?

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/streams/StreamsServiceImpl.java
```diff
@@ -73,7 +73,7 @@ private Stream createStreamFromProperties(String streamName, Map<String, String>
                     streamBean.setDescription(value);
                     break;
                 case "obr":
-                    streamBean.setObrLocation(value);
+                    streamBean.setObrs(value);
```
techcobweb: Could this be called setObrsFromCommaSeparatedList

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/streams/IStream.java
```diff
@@ -5,18 +5,23 @@
...
     String getDescription();
 
-    String getMavenRepositoryUrl();
+    URL getMavenRepositoryUrl();
 
-    String getTestCatalogUrl();
+    URL getTestCatalogUrl();
 
-    String getObrLocation();
+    List<String> getObrs();
```
techcobweb: Can we have an OBR object ?

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/streams/IStream.java
```diff
@@ -5,18 +5,23 @@
  */
 package dev.galasa.framework.spi.streams;
 
+import java.net.URL;
+import java.util.List;
+
 public interface IStream {
 
     String getName();
```
techcobweb, line 13: Is this name validated already by the time we have an IStream object ?

modules/framework/galasa-parent/dev.galasa.framework/src/testFixtures/java/dev/galasa/framework/mocks/MockStream.java
```diff
@@ -67,19 +58,28 @@ public void setDescription(String description) {
...
+        this.obrs = obrs;
     }
 
     public void setEnabled(boolean isEnabled) {
         this.isEnabled = isEnabled;
     }
+
+    @Override
+    public boolean isValid() {
+        boolean isValid = (
```
techcobweb: implementation inside a test class ?

modules/framework/galasa-parent/galasa-boot/src/main/java/dev/galasa/boot/Launcher.java
```diff
@@ -172,7 +171,7 @@ protected void launch(String[] args) throws LauncherException, InterruptedExcept
                 felixFramework.runTest(bootstrapProperties, overridesProperties);
             } else if (isResourceManagement) {
                 logger.debug("Resource Management");
-                MonitorConfiguration monitorConfig = getResourceMonitorConfigFromEnvironment(env);
+                MonitorConfiguration monitorConfig = new MonitorConfiguration(env);
```
techcobweb: Should this be called ResourceManagementConfiguration ?

techcobweb (approved):

modules/framework/galasa-parent/dev.galasa.framework.api.streams/src/main/java/dev/galasa/framework/api/streams/internal/routes/StreamsTransform.java
eamansour: Addressed in https://github.com/galasa-dev/galasa/pull/236/commits/a9ef197a47c1c7da97f4c1566f80ca24bce91184

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/MonitorFilter.java
eamansour: Renamed in https://github.com/galasa-dev/galasa/pull/236/commits/a9ef197a47c1c7da97f4c1566f80ca24bce91184

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/ResourceManagementProviders.java
eamansour: Addressed in https://github.com/galasa-dev/galasa/pull/236/commits/a9ef197a47c1c7da97f4c1566f80ca24bce91184

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/ResourceManagement.java
eamansour: Addressed in https://github.com/galasa-dev/galasa/pull/236/commits/a9ef197a47c1c7da97f4c1566f80ca24bce91184

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/MonitorFilter.java
eamansour: Addressed in https://github.com/galasa-dev/galasa/pull/236/commits/a9ef197a47c1c7da97f4c1566f80ca24bce91184

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/streams/Stream.java
eamansour: Addressed in https://github.com/galasa-dev/galasa/pull/236/commits/a9ef197a47c1c7da97f4c1566f80ca24bce91184

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/streams/Stream.java
eamansour: Addressed in https://github.com/galasa-dev/galasa/pull/236/commits/a9ef197a47c1c7da97f4c1566f80ca24bce91184

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/streams/Stream.java
eamansour: Addressed in https://github.com/galasa-dev/galasa/pull/236/commits/a9ef197a47c1c7da97f4c1566f80ca24bce91184

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/streams/StreamsServiceImpl.java
eamansour: Addressed in https://github.com/galasa-dev/galasa/pull/236/commits/a9ef197a47c1c7da97f4c1566f80ca24bce91184

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/streams/IStream.java
eamansour: Addressed in https://github.com/galasa-dev/galasa/pull/236/commits/a9ef197a47c1c7da97f4c1566f80ca24bce91184

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/streams/IStream.java
eamansour: Addressed in https://github.com/galasa-dev/galasa/pull/236/commits/a9ef197a47c1c7da97f4c1566f80ca24bce91184

modules/framework/galasa-parent/dev.galasa.framework/src/testFixtures/java/dev/galasa/framework/mocks/MockStream.java
eamansour: Addressed in https://github.com/galasa-dev/galasa/pull/236/commits/a9ef197a47c1c7da97f4c1566f80ca24bce91184

modules/framework/galasa-parent/galasa-boot/src/main/java/dev/galasa/boot/Launcher.java
eamansour: Renamed in https://github.com/galasa-dev/galasa/pull/236/commits/a9ef197a47c1c7da97f4c1566f80ca24bce91184

PR #237, opened 2025-03-26
Add REST APIs to get monitors from Kubernetes deployments

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2067

## Changes
- Added `GET /monitors` and `GET /monitors/{monitor-name}` endpoints to get monitors from their associated Kubernetes deployments that are installed when the Galasa service is installed

PR #238, opened 2025-03-26
Add REST API to scale resource monitor deployments in Kubernetes

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2067

Note: This PR builds from changes in https://github.com/galasa-dev/galasa/pull/237 - once that PR is delivered, this branch will be rebased to remove the duplicate changes and make reviewing this PR easier.

## Changes
- Added `PUT /monitors/{monitor-name}` REST API endpoint that can be used as shown below to update the number of replicas for the given monitor's Kubernetes deployment to 0 or 1 based on the boolean value provided in the `isEnabled` request body parameter:
    ```json
    > PUT /monitors/my-custom-monitor

    {
        "data": {
            "isEnabled": true
        }
    }

    > Response: 200 with the updated deployment JSON
    ```

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
+++ b/modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
@@ -1202,7 +1202,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/GalasaProperty'
+              $ref: '#/components/schemas/UpdateGalasaMonitorRequest'
       tags:
         - Monitors API
       responses:
--- a/modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
+++ b/modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
@@ -1202,12 +1202,14 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/UpdateGalasaMonitorRequest'
+              $ref: '#/components/schemas/GalasaMonitor'
       tags:
         - Monitors API
       responses:
         '204':
           description: The Galasa Monitor was successfully enabled or disabled
+        '400':
+          $ref: "#/components/responses/BadRequest"
         '401':
           $ref: "#/components/responses/Unauthorized"
         '403':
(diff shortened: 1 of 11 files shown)


techcobweb:
Does the `galasactl resources apply -f` allow you to set the enabled/disabled state ?
I guess we need a separate story for that...

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -2684,6 +2822,12 @@ components:
           type: object
         trace:
           type: boolean
+    UpdateGalasaMonitorRequest:
```
techcobweb, line 2827: I think this should mirror the definition of the monitor from the Get call more I think.

eg: Put the isEnabled in a 'data' section ?

When we have the monitor definition in couchdb and allow more things to be set (description, includes/excludes) we will want this structure built out more, so the more it looks like the output of Get the better ? 

I guess it's subjective, but... 

Thoughts ?

techcobweb (approved):

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
techcobweb: OR maybe we should PUT the same structure, and ignore the fields which aren't changeable ? Then it would match the apply -f payload more ?
(Not sure I like that as it will get confusing which fields are settable and which are not).

techcobweb:
Shouldn't the PUT return the updated definition of the monitor, on a 200 rather than a 204 ?

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
eamansour: Sounds good - addressed in https://github.com/galasa-dev/galasa/pull/238/commits/05beb9686ac3720b9ad24b2e4edc6e5d35827503 and https://github.com/galasa-dev/galasa/pull/238/commits/d5454ba4865b525f4af52946c6419ee89b94504a

eamansour:
> Shouldn't the PUT return the updated definition of the monitor, on a 200 rather than a 204 ?

Done in https://github.com/galasa-dev/galasa/pull/238/commits/05beb9686ac3720b9ad24b2e4edc6e5d35827503

> Does the `galasactl resources apply -f` allow you to set the enabled/disabled state ? I guess we need a separate story for that...

Not yet, that will be available with in phase 2 where users will be able to set monitor definitions.

PR #239, opened 2025-03-27
Iss1920 - Ability to run custom code in test class when failures encountered.

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/1920

- New annotation `TestResultProvider` to inject a `TestResultProvider` into a test case, to access the result in between the Test and AfterClass methods, to run custom code only in the event of a test failure.

- The TestClassWrapper and TestRunner now inform the active managers about the "result so far" of a test case, instead of only making the result available after the lifecycle of the test. 

- Unit tests for this change set in CoreManagerImplTest.

- Galasa test that exercises this change in TestTestResultProvider - but perhaps not a good idea to deliver this with the `@Test` annotation as it will always fail and report a failure in our IVTs stream. Happy to remove for this reason.


modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
```diff
@@ -228,33 +230,35 @@ public void runTestMethods(@NotNull ITestRunManagers managers, IDynamicStatusSto
...
         for (GenericMethodWrapper afterClassMethod : afterClassMethods) {
             afterClassMethod.invoke(managers, this.testClassObject, null);
             if (afterClassMethod.fullStop()) {
-                if (this.result == null) {
-                    this.result = Result.failed("AfterClass method failed");
+                if (getResult() == null) {
+                    setResult( Result.failed("AfterClass method failed") , managers );
                 }
             }
         }
```
eamansour: Would it be worth splitting parts of this method out into smaller methods? The bits to run the `@BeforeClass` and `@AfterClass` are very similar, and the bit that runs the `@Test` methods could also be separated?

(I know this code was already there, just thinking about tidying it up!)

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
```diff
@@ -228,33 +230,35 @@ public void runTestMethods(@NotNull ITestRunManagers managers, IDynamicStatusSto
...
         // Test result
-        logger.info(LOG_ENDING + LOG_START_LINE + LOG_ASTERS + LOG_START_LINE + "*** " + this.result.getName()
+        logger.info(LOG_ENDING + LOG_START_LINE + LOG_ASTERS + LOG_START_LINE + "*** " + getResult().getName()
         + " - Test class " + testClass.getName() + LOG_START_LINE + LOG_ASTERS);
 
-        this.testStructure.setResult(this.result.getName());
+        this.testStructure.setResult(getResult().getName());
 
-        managers.testClassResult(this.result, null);
+        managers.testClassResult(getResult(), null);
```
eamansour, line 215: Really minor - do we need to call `getResult()` everywhere if we could instead use `this.resultData`?

eamansour (approved):
Minor comments but approved

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
jadecarino: We went with a getter method here as Mike suggested that was a better way of coding it - I believe for unit testing purposes.

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
jadecarino: Thanks Eamonn - I've now refactored this method and split the running of the BeforeClass/AfterClass methods and running of Test methods into their own methods.

eamansour (approved):

PR #242, opened 2025-03-31
protect against trailing slashes in the streams repo path

techcobweb:

# Why ?
If the repository part of the streams definition contains a URL which ends in a slash, none o f the tests can load, as the loading code creates a URL with '//' (double-slash) in the path, which causes the load code to hang.

Solution: Be more careful when constructing the URL path for a test class we want to load
and avoid double-slashes in a path into the maven repo location.


aashir21 (approved):
👍🏻

PR #243, opened 2025-04-01
mcobbett test pod mem constraints

techcobweb:
# Why ?

See [Kubernetes pods should be launched with a min and max memory constraint.
#2164](https://github.com/galasa-dev/projectmanagement/issues/2164)

Added 4 settings in the configmap which the engine controller looks at:

- engine_memory_request - The amount of memory requested for the test pod. In Megabytes.
- engine_memory_limit - The memory limit of the pod, after which it errors. In Megabytes.
- engine_cpu_request - The amount of CPU (in thousandths of a CPU) we request for the test pod A 0 value means no value is set.
- engine_cpu_limit - The max amount of CPU the test can consume. A 0 value means no limit gets set.

Also protected the resource manager code from a null pointer when it shuts down and there is no etcd watcher set up yet.

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
@@ -340,6 +340,7 @@ public class TestPodScheduler implements Runnable {
         return tolerationsList;
     }
 
+
     private V1Container createTestContainer(String runName, String engineName, boolean isTraceEnabled) {
         V1Container container = new V1Container();
         container.setName("engine");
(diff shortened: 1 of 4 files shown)


modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
```diff
@@ -425,7 +436,7 @@ private List<V1EnvVar> createTestContainerEnvVariables() {
         // "galasa_maven_helper_repo"));
         //
         // envs.add(createValueEnv("GALASA_ENGINE_TYPE", engineLabel));
-        envs.add(createValueEnv("MAX_HEAP", Integer.toString(this.settings.getEngineMemory()) + "m"));
+        envs.add(createValueEnv("MAX_HEAP", Integer.toString(this.settings.getEngineMemoryHeapSizeMegabytes()) + "m"));
```
eamansour, line 453: Possibly 'Mi' for megabytes?

eamansour (approved):

modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
techcobweb: For `java -Xmx` The 'm' means megabytes

PR #244, opened 2025-04-01
Cache IMS TM version and applid properties per tag

ideeley:
https://github.com/galasa-dev/projectmanagement/issues/2177

ImstmProperties was previously caching single instances of the imstm.dse.tag.[TAG].version and imstm.dse.tag.[TAG].applid properties. These are now cached in maps with the tag as the key.


jadecarino (approved):
Looks good to me

PR #248, opened 2025-04-08
Delete run DSS properties for local runs only and delete test pods for finished runs

eamansour:
## Why?
Relates to https://github.com/galasa-dev/projectmanagement/issues/2181

Test runs are being removed from the DSS too quickly for clients to track, which has been causing issues in various areas of Galasa. By deleting a run from the DSS while the resource monitor is down, there is also a risk that resources provisioned for that run do not get cleaned up properly when the resource monitor comes back up - this PR reverts to keeping DSS properties for a run around so that the resource monitor can clean them up after a certain period of time.

Keeping the DSS properties does mean that the engine controller would keep the Kubernetes pods associated with runs around until the resource monitor deletes the DSS properties for the runs, which in turn triggers the engine controller's "deleted runs" scan to delete the pods. To avoid a build-up of completed test pods in Kubernetes, the engine controller now deletes pods for runs that have finished.

## Changes
- The test runner no longer deletes the run's DSS properties at the end of a run, the deletion of a run's DSS properties is now handled by the resource monitor
- The engine controller thread that cleans up Kubernetes pods for test runs now deletes pods for runs that have finished as well as runs that no longer exist in the DSS (same as it was before)

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/RunPodCleanup.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/RunPodCleanup.java
@@ -11,6 +11,7 @@ import java.util.Map;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import dev.galasa.framework.TestRunLifecycleStatus;
 import dev.galasa.framework.spi.IFrameworkRuns;
 import dev.galasa.framework.spi.IRun;
 import io.kubernetes.client.openapi.ApiException;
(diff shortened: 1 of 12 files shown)


modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/RunPodCleanup.java
```diff
@@ -0,0 +1,95 @@
+/*
```
eamansour, line 1: Not a new file, just renamed from RunDeleted to RunPodCleanup - compare with https://github.com/galasa-dev/galasa/pull/248/files#diff-c3d7fa558add0e5375a2b6e5f93412b1a5cb3b2d3a738135c8bc68d13e1edde5

techcobweb (approved):

PR #256, opened 2025-04-11
Fix ContinueOnTestFailure annotation not behaving as expected

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2193

Previously, the `@ContinueOnTestFailure` annotation was defined but wasn't used anywhere in the Galasa code, effectively rendering it useless. Now, the test class wrapper checks whether the annotation is provided, and if it is, the test class will continue to run regardless of any failures and regardless of what the `framework.continue.on.test.failure` CPS property is set to.

## Changes
- When a test class is annotated with `@ContinueOnTestFailure` and the CPS property `framework.continue.on.test.failure` is set to `false`, the test class will now correctly continue to run regardless of test failures

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework/src/testFixtures/java/dev/galasa/framework/mocks/MockIConfigurationPropertyStoreService.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework/src/testFixtures/java/dev/galasa/framework/mocks/MockIConfigurationPropertyStoreService.java
@@ -27,7 +27,7 @@ public class MockIConfigurationPropertyStoreService implements IConfigurationPro
     @Override
     public @Null String getProperty(@NotNull String prefix, @NotNull String suffix, String... infixes)
             throws ConfigurationPropertyStoreException {
-        String propertyKey = prefix + "." + suffix;;
+        String propertyKey = prefix + "." + suffix;
         if (infixes.length > 0) {
             propertyKey = prefix + "." + String.join(".", infixes) + "." + suffix;
         }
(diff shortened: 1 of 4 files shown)


modules/framework/galasa-parent/dev.galasa.framework/src/testFixtures/java/dev/galasa/framework/mocks/MockIConfigurationPropertyStoreService.java
```diff
@@ -27,7 +27,11 @@ public MockIConfigurationPropertyStoreService() {
     @Override
     public @Null String getProperty(@NotNull String prefix, @NotNull String suffix, String... infixes)
             throws ConfigurationPropertyStoreException {
-        return null ;
+        String propertyKey = prefix + "." + suffix;;
```
jadecarino: Super minor I know it doesn't break anything but there's a double ;

jadecarino (approved):
Super minor comment but approved! Thanks for implementing this

modules/framework/galasa-parent/dev.galasa.framework/src/testFixtures/java/dev/galasa/framework/mocks/MockIConfigurationPropertyStoreService.java
eamansour: Good spot, thanks - removed in https://github.com/galasa-dev/galasa/pull/256/commits/6e87817ea6271d7b5f5dc1f4567862672fd80118

PR #272, opened 2025-04-17
Added integration tests for streams commands

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/1614

### DIFF
--- a/modules/cli/pkg/streams/streamsGet_test.go
+++ b/modules/cli/pkg/streams/streamsGet_test.go
@@ -268,7 +268,7 @@ func TestCanGetAStreamByNameWhenStreamExistsGivesUnexpectedErrorSummaryFormat(t 
 	assert.Contains(t, err.Error(), "GAL1238E")
 }
 
-func TestCanGetAStreamByNameWhenStreamExistsGivesUnexpectedErrorSummaryFormat2(t *testing.T) {
+func TestCanGetAStreamByNameWhenStreamExistsWithRequestBodyGivesUnexpectedErrorSummaryFormat(t *testing.T) {
 	// Given...
 	outputFormat := "summary"
 
--- a/modules/cli/pkg/streams/streamsGet_test.go
+++ b/modules/cli/pkg/streams/streamsGet_test.go
@@ -267,3 +267,68 @@ func TestCanGetAStreamByNameWhenStreamExistsGivesUnexpectedErrorSummaryFormat(t 
 	assert.NotNil(t, err, "GetStreams returned an error when none was expected")
 	assert.Contains(t, err.Error(), "GAL1238E")
 }
+
+func TestCanGetAStreamByNameWhenStreamExistsGivesUnexpectedErrorSummaryFormat2(t *testing.T) {
+	// Given...
+	outputFormat := "summary"
+
+	body := `
+{
+  "apiVersion": "galasa-dev/v1alpha1",
+  "kind": "GalasaStream",
+  "metadata": {
+    "name": "mystream",
+    "url": "http://localhost:8080/api/streams/myStream",
+    "description": "A stream which I use to..."
+  },
+  "data": {
+    "isEnabled": true,
+    "repository": {
+      "url": "http://mymavenrepo.host/testmaterial"
+    },
+    "obrs": [
+      {
+        "groupId": "com.ibm.zosadk.k8s",
+        "artifactId": "com.ibm.zosadk.k8s.obr",
+        "version": "0.1.0-SNAPSHOT"
+      }
+    ],
+    "testCatalog": {
+      "url": "http://mymavenrepo.host/testmaterial/com.ibm.zosadk.k8s/com.ibm.zosadk.k8s.obr/0.1.0-SNAPSHOT/testcatalog.yaml"
+    }
+  }
+}
+`
+
+	// Create the expected HTTP interactions with the API server.
+	getStreamInteraction := utils.NewHttpInteraction("/streams/mystream", http.MethodGet)
+	getStreamInteraction.WriteHttpResponseFunc = func(writer http.ResponseWriter, req *http.Request) {
+		writer.Header().Set("Content-Type", "application/json")
+		writer.WriteHeader(http.StatusInternalServerError)
+		writer.Write([]byte(body))
+	}
+
+	interactions := []utils.HttpInteraction{
+		getStreamInteraction,
+	}
+
+	server := utils.NewMockHttpServer(t, interactions)
+	defer server.Server.Close()
+
+	console := utils.NewMockConsole()
+	apiServerUrl := server.Server.URL
+	apiClient := api.InitialiseAPI(apiServerUrl)
+	mockByteReader := utils.NewMockByteReader()
+
+	// When...
+	err := GetStreams(
+		"mystream",
+		outputFormat,
+		apiClient,
+		console,
+		mockByteReader)
+
+	// Then...
+	assert.NotNil(t, err, "GetStreams returned an error when none was expected")
+	assert.Contains(t, err.Error(), "GAL1239E")
+}
--- a/modules/cli/test-scripts/streams-tests.sh
+++ b/modules/cli/test-scripts/streams-tests.sh
@@ -114,31 +114,89 @@ function configure_stream {
     --bootstrap $bootstrap \
     --log -"
 
-    info "Command is: $cmd1"
-    info "Command is: $cmd2"
-    info "Command is: $cmd3"
-    info "Command is: $cmd4"
+    info "Adding description property: $cmd1"
+    info "Adding maven repo url property: $cmd2"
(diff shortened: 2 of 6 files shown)


modules/cli/pkg/streams/streamsGet_test.go
```diff
@@ -267,3 +267,68 @@ func TestCanGetAStreamByNameWhenStreamExistsGivesUnexpectedErrorSummaryFormat(t
 	assert.NotNil(t, err, "GetStreams returned an error when none was expected")
 	assert.Contains(t, err.Error(), "GAL1238E")
 }
+
+func TestCanGetAStreamByNameWhenStreamExistsGivesUnexpectedErrorSummaryFormat2(t *testing.T) {
```
eamansour: Is this test name accurate? The 2 at the end seems a bit strange

modules/cli/test-scripts/streams-tests.sh
```diff
@@ -0,0 +1,250 @@
...
+    local cmd4="${BINARY_LOCATION} properties set --namespace framework \
+    --name $obr_prop_name \
+    --value mvn:myorg/myartifact/0.0.1/obr \
+    --bootstrap $bootstrap \
+    --log -"
+
+    info "Command is: $cmd1"
+    info "Command is: $cmd2"
+    info "Command is: $cmd3"
+    info "Command is: $cmd4"
```
eamansour: Maybe it would be better to log a command then run it, otherwise it won't be clear which command failed if an error occurs

modules/cli/test-scripts/streams-tests.sh
```diff
@@ -0,0 +1,250 @@
...
+
+    $cmd1
+    $cmd2
+    $cmd3
+    $cmd4
+    local rc=$?
```
eamansour: This `rc=$?` only checks the return code of `$cmd4`, so if any of the others fail, we won't exit.

modules/cli/test-scripts/streams-tests.sh
```diff
@@ -0,0 +1,250 @@
...
+
+    # We expect a return code of 0 because this is a properly formed properties set command.
+    if [[ "${rc}" != "0" ]]; then 
+        error "Failed to create property with name and value used."
+        return 1
+    fi
+    
+    return 0
```
eamansour: No need to return 0

modules/cli/test-scripts/streams-tests.sh
```diff
@@ -0,0 +1,250 @@
...
+}
+
+#-----------------------------------------------------------------------------------------
+# Tests
+#----------------------------------------------------------------------------------------- 
+function streams_create {
+    h2 "Performing properties set with name and value parameter used: create..."
```
eamansour: This message could be clarified to say that we're creating a test stream for these tests?

modules/cli/test-scripts/streams-tests.sh
```diff
@@ -0,0 +1,250 @@
...
+    --log -"
+
+    info "Command is: $cmd"
+
+    output_file="$ORIGINAL_DIR/temp/streams-get-output.txt"
+    $cmd | tee $output_file
+    if [[ "${rc}" != "0" ]]; then 
+        error "Failed to get stream with name used: command failed."
+        exit 1
+    fi
```
eamansour: I would separate the `streams get` call out into a separate function as this is a test

modules/cli/test-scripts/streams-tests.sh
```diff
@@ -0,0 +1,250 @@
...
+
+    # Check that the previous properties set created a stream property
+    cat $output_file | grep "$stream_name" -q
+
+    rc=$?
+    # We expect a return code of 0 because this is a properly formed streams get command.
+    if [[ "${rc}" != "0" ]]; then 
+        error "Failed to create stream with name and value used."
+        exit 1
+    fi
```
eamansour: Failed to get stream? Or something saying that the output is incorrect/missing the stream name?

modules/cli/test-scripts/streams-tests.sh
```diff
@@ -0,0 +1,250 @@
...
+    cat $output_file | grep "Galasa" -q -E
+    cat $output_file | grep "mystream" -q -E
+    cat $output_file | grep "enabled" -q -E
+
+    rc=$?
+    # We expect a return code of 0 because this is a properly formed properties get command.
+    if [[ "${rc}" != "0" ]]; then 
+        error "Stream successfully created, but value incorrect."
+        exit 1
+    fi
```
eamansour: This rc=$? is only checking the `cat $output_file | grep "enabled" -q -E` line's return code. The other return codes are ignored

modules/cli/test-scripts/streams-tests.sh
```diff
@@ -0,0 +1,250 @@
...
+
+    success "Stream set with name and value used seems to have been created correctly."
+}
```
eamansour, line 316: I think we need more tests to check that `streams get` works as expected.

We could add a test for a simple `streams get` without passing in a `--name`, a test for `streams get --name`, pass in `--format yaml`, a test for `streams get` with a `--name` for a stream that doesn't exist?

modules/cli/test-scripts/streams-tests.sh
```diff
@@ -0,0 +1,250 @@
...
+    info "Command is: $cmd"
+
+    output_file="$ORIGINAL_DIR/temp/streams-get-output.txt"
+    $cmd | tee $output_file
+
+    # Check for successful deletion, it should return 404 response code
+    cat $output_file | grep "GAL5420E: Unable to retrieve a stream with the given stream name. No such stream exists." -q -E
+    
+
+    success "Properties delete with name used seems to have been deleted correctly."
```
eamansour: Streams delete

modules/cli/test-scripts/streams-tests.sh
```diff
@@ -114,53 +114,106 @@ function configure_stream {
...
+    if [[ "${rc}" != "0" ]]; then 
+        error "Failed to get streams: command failed."
+        exit 1
+    fi
+
+    success "Fetched all streams"
+}
+
+function streams_create {
+    h2 "Creating a test stream for get test"
```
eamansour: This seems to be a duplicate message from the streams get tests?

modules/cli/test-scripts/streams-tests.sh
```diff
@@ -172,23 +225,109 @@ function streams_create {
...
+
+    # Check if any of the greps failed
+    if [[ "$rc1" != "0" || "$rc2" != "0" || "$rc3" != "0" ]]; then
+        # Report which specific check failed
+        [[ "$rc1" != "0" ]] && error "Missing 'Galasa' in stream output"
+        [[ "$rc2" != "0" ]] && error "Missing '$stream_name' in stream output"
+        [[ "$rc3" != "0" ]] && error "Missing 'enabled' in stream output"
+        error "Stream successfully created, but values incorrect."
+        exit 1
+    fi
```
eamansour: We could check for errors directly after running each command instead of having to figure out which command failed?

modules/cli/test-scripts/streams-tests.sh
```diff
@@ -172,23 +225,109 @@ function streams_create {
...
 
+    success "Stream set with name and value used seems to have been created correctly."
+}
+
+function streams_get_yaml_format {
+    h2 "Creating a test stream for get test"
```
eamansour: Looks like the same message as the streams get tests, would be good to differentiate between them so that it's easier to figure out where things fail if errors pop up

modules/cli/test-scripts/streams-tests.sh
```diff
@@ -172,23 +225,109 @@ function streams_create {
...
+
+    # Check if any of the greps failed
+    if [[ "$rc1" != "0" || "$rc2" != "0" || "$rc3" != "0" ]]; then
+        # Report which specific check failed
+        [[ "$rc1" != "0" ]] && error "Missing 'Galasa' in stream output"
+        [[ "$rc2" != "0" ]] && error "Missing '$stream_name' in stream output"
+        [[ "$rc3" != "0" ]] && error "Missing 'GalasaStream' in stream output"
+        error "Stream successfully created, but values incorrect."
+        exit 1
+    fi
```
eamansour: Same here - we could check for errors directly after running each command instead of having to figure out which command failed?

modules/cli/test-scripts/streams-tests.sh
```diff
@@ -172,23 +225,109 @@ function streams_create {
...
+
+    success "Stream set with name and value used seems to have been created correctly and formatted in YAML"
+}
+
+function streams_get_missing_name {
+    h2 "Creating a test stream for get test"
```
eamansour: Again, would be good to say what this is testing

eamansour (approved):

PR #276, opened 2025-04-28
Update run RAS records when runs are requeued

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2209

## Changes
- Requeuing runs now correctly updates the RAS records of past runs, marking runs as "finished(Requeued)", while allowing re-runs to continue by removing the `interruptReason` DSS property and adding an action to the `rasActions` DSS property
- For runs that are detected by the dead heartbeat monitor, the monitor will interrupt the runs to be marked as "finished(Hung)" (if it decides to kill the run), or "finished(Requeued)" (if it decides to re-queue the run)
- At the end of a test pod's execution (in the case of a normal run that was re-queued due to a lack of available resources), the pod will process its own RAS actions before having its properties deleted from the DSS


modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/test/java/dev/galasa/framework/api/ras/internal/routes/TestResultNamesRoute.java
```diff
@@ -83,6 +83,8 @@ private String generateExpectedJSON (List<IRunResult> mockInputRunResults, boole
 		resultNames.add("EnvFail");
 		resultNames.add("Ignored");
 		resultNames.add("Cancelled");
+		resultNames.add("Requeued");
```
techcobweb, line 86: Wondering if the same data is in the CLI or whether it doesn't restrict inputs from users in terms of state queries.

modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/InterruptedRunEventProcessor.java
```diff
@@ -59,8 +60,22 @@ public void run() {
...
-                    markRunFinishedInDss(interruptEvent);
-                    processRasActions(interruptEvent);
+                    String runName = interruptEvent.getRunName();
+                    List<RunRasAction> rasActions = interruptEvent.getRasActions();
+                    rasActionProcessor.processRasActions(runName, rasActions);
+                    
+                    String interruptReason = interruptEvent.getInterruptReason();
+                    switch (interruptReason) {
+                        case Result.CANCELLED:
+                        case Result.HUNG:
```
techcobweb, line 70: What decides that it's hung ?

techcobweb (approved):

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/test/java/dev/galasa/framework/api/ras/internal/routes/TestResultNamesRoute.java
eamansour: The CLI restricts the results passed in via `runs get` to these values. When passing in something that isn't a known result, it responds with this error:
```
❯ galasactl runs get --age 1d --result bob
Error: GAL1087E: Invalid '--result' parameter value: 'bob'. The possible result values currently in the Ecosystem Result Archive Store (RAS) are: 'Cancelled', 'EnvFail', 'Failed', 'Hung', 'Ignored', 'Passed', 'Requeued
```

modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/InterruptedRunEventProcessor.java
eamansour: Currently, only the dead heartbeat monitor sets this `Hung` result (see changes in https://github.com/galasa-dev/galasa/pull/276/files#diff-aaf2f845a8022ce883024fcd2a174b2604a61f0cdedf009c742cf0e30702d8b8).

PR #277, opened 2025-04-30
Add trace logging to 3270 network communications

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2188

## Changes
- Added extra trace messages to the 3270 manager to help diagnose the defect described in the story above. These messages will only appear when `--trace` is passed in when running tests.


techcobweb:
I would think that dev.galasa.zos3270.internal.comms.TrustAllCerts should be covered in trace statements, then we can see what it's being asked to do... ?

techcobweb (approved):

PR #282, opened 2025-05-01
RAS Runs now return method details

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2202


modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -137,9 +139,16 @@ private String retrieveResults(RasQueryParameters queryParams) throws InternalSe
 
         RasRunResultPage runsPage = null;
         String responseJson = null;
+
+		boolean isMethodDetailsRequested = false;
+		String detailParam = queryParams.getDetail();
+		if(detailParam != null && !detailParam.isEmpty() && detailParam.equals("methods")) {
+			isMethodDetailsRequested = true;
+		}
```
eamansour: If more details are added later, this would become a bit messy with having to pass around lots of boolean flags.

We could encapsulate all the detail selections into a separate class and pass that around instead? Then we can use setters to turn on these flags?

Just a suggestion, I'm happy to leave this with you to decide if you want to make more changes or keep this as-is.

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -508,6 +526,10 @@ private Comparator<RasRunResult> buildRunsComparator(RasQueryParameters queryPar
         return runsComparator;
     }
 
+	private boolean isTestRunShort(boolean isMethodDetailsRequested) {
+		return isMethodDetailsRequested ? false : true;
+	}
```
eamansour: Minor - we could just say `!isMethodDetailsRequested` instead of making a function call?

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -331,14 +347,15 @@ private List<RasRunResult> getRuns(List<IRasSearchCriteria> critList) throws Res
 		IRasSearchCriteria[] criteria = new IRasSearchCriteria[critList.size()];
 
 		critList.toArray(criteria);
-
+		boolean isMethodDetailsRequested = critList.stream().anyMatch(e -> e instanceof RasSearchCriteriaDetail);
```
eamansour: This operation is more expensive than passing in the boolean flag/detail settings.

Don't think we need RasSearchCriteriaDetail

modules/extensions/galasa-extensions-parent/dev.galasa.ras.couchdb/src/main/java/dev/galasa/ras/couchdb/internal/CouchdbDirectoryService.java
```diff
@@ -506,6 +507,9 @@ private JsonObject buildGetRunsQuery(IRasSearchCriteria... searchCriterias) thro
             } else if (searchCriteria instanceof RasSearchCriteriaStatus) {
                 RasSearchCriteriaStatus sStatus = (RasSearchCriteriaStatus) searchCriteria;
                 inArray(and, "status", sStatus.getStatusesAsStrings());
+            } else if (searchCriteria instanceof RasSearchCriteriaDetail) {
+                RasSearchCriteriaDetail sDetail = (RasSearchCriteriaDetail) searchCriteria;
+                inArray(and, "detail", sDetail.getDetails());
```
eamansour: `detail` isn't a field in the test structure. Not sure if this is needed...

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -266,6 +278,10 @@ private List<IRasSearchCriteria> getCriteria(RasQueryParameters queryParams) thr
 			RasSearchCriteriaSubmissionId submissionIdCriteria = new RasSearchCriteriaSubmissionId(submissionId);
 			critList.add(submissionIdCriteria);
 		}
+		if (detail != null && !detail.isEmpty()) {
+			RasSearchCriteriaDetail detailCriteria = new RasSearchCriteriaDetail(detail);
+			critList.add(detailCriteria);
+		}
```
eamansour: Likely not needed since the detail parameter controls what additional information is pulled out from the test structure. We won't be searching for runs based on the `detail` parameter.

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/ras/RasSearchCriteriaDetail.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package dev.galasa.framework.spi.ras;
+
+import javax.validation.constraints.NotNull;
+
+import dev.galasa.framework.spi.teststructure.TestStructure;
+
+public class RasSearchCriteriaDetail implements IRasSearchCriteria {
```
eamansour: See comments above, I don't think this is needed.

modules/extensions/galasa-extensions-parent/dev.galasa.ras.couchdb/src/main/java/dev/galasa/ras/couchdb/internal/CouchdbDirectoryService.java
```diff
@@ -40,6 +40,7 @@
 import dev.galasa.framework.spi.ras.IRasSearchCriteria;
 import dev.galasa.framework.spi.ras.RasRunResultPage;
 import dev.galasa.framework.spi.ras.RasSearchCriteriaBundle;
+import dev.galasa.framework.spi.ras.RasSearchCriteriaDetail;
```
eamansour: Needs removing?

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/common/RasDetailsQueryParams.java
```diff
@@ -0,0 +1,28 @@
...
+        
+    }
+
+    public RasDetailsQueryParams(String detail) {
+        this.detail = detail;
+    }
+
+    public String getDetail() {
+        return detail;
+    }
```
eamansour: Instead of getDetail, maybe we could have a method like `isMethodDetailsRequested()` that returns true if "methods" has been passed into the detail param? That way, we can ask this class if we need to include method details or not

(feel free to think of a better method name!)

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/common/RasDetailsQueryParams.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package dev.galasa.framework.api.ras.internal.common;
+
+public class RasDetailsQueryParams {
+    
+    private String detail;
```
eamansour: I'm wondering if this should be a `Set<String>` that we populate in the constructor using the query param's string value rather than a string

Probably fine for now since we only have "methods" - happy to keep as-is :)

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -81,9 +82,11 @@ public class RunQueryRoute extends RunsRoute {
 	public static final String QUERY_PARAMETER_CURSOR = "cursor";
 	public static final String QUERY_PARAMETER_RUNNAME = "runname";
 	public static final String QUERY_PARAMETER_RUNID = "runid";
+	public static final String QUERY_PARAMATER_DETAIL = "detail";
+	public static final String DETAILS_METHOD_QUERY_PARAM_VALUE = "methods";
```
eamansour: Perhaps move the methods constant into the DetailsQueryParam class?

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -92,6 +95,7 @@ public class RunQueryRoute extends RunsRoute {
 
 
 	private static final GalasaGson gson = new GalasaGson();
+	private final RasDetailsQueryParams rasDetailsQueryParams;
```
eamansour: This shouldn't be a field since the query params will vary per request and this could potentially cause some issues when multiple requests come in at the same time

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -279,8 +277,7 @@ private List<IRasSearchCriteria> getCriteria(RasQueryParameters queryParams) thr
 			critList.add(submissionIdCriteria);
 		}
 		if (detail != null && !detail.isEmpty()) {
-			RasSearchCriteriaDetail detailCriteria = new RasSearchCriteriaDetail(detail);
-			critList.add(detailCriteria);
+			rasDetailsQueryParams.setDetail(detail);
```
eamansour: I think this should be done outside this `getCriteria` method since the rest of the method is adding to a list of search criteria, whereas this is setting a detail

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -526,8 +520,14 @@ private Comparator<RasRunResult> buildRunsComparator(RasQueryParameters queryPar
...
-	private boolean isTestRunShort(boolean isMethodDetailsRequested) {
-		return isMethodDetailsRequested ? false : true;
+	private boolean isTestRunShort(String detailParam) {
+		
+		boolean isShort = true;
+		if(detailParam != null && detailParam.equals(DETAILS_METHOD_QUERY_PARAM_VALUE)) {
+			isShort = false;
+		}
+
+		return isShort;
```
eamansour: I think this won't be needed if we add a method inside the RasDetailsQueryParams class that returns true if we've passed in "methods" into the details param.

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -392,8 +386,8 @@ private List<IRunResult> getRunsByRunName(String runName) throws ResultArchiveSt
 		return runs;
 	}
 
-    private List<RasRunResult> convertRunsToRunResults(List<IRunResult> runs,boolean isMethodDetailsRequested) throws ResultArchiveStoreException {
-		boolean isShort = isTestRunShort(isMethodDetailsRequested);
+    private List<RasRunResult> convertRunsToRunResults(List<IRunResult> runs) throws ResultArchiveStoreException {
+		boolean isShort = isTestRunShort(rasDetailsQueryParams.getDetail());
```
eamansour: isShort and isTestRunShort could confuse people - perhaps this should be something like isMethodDetailsExcluded? Or something better?

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -196,9 +193,10 @@ private RasSortField formatSortField(RasSortField sortValue) {
         return sortField;
     }
 
-    private List<RasRunResult> getRunsByIds(List<String> runIds, boolean isMethodDetailsRequested) throws InternalServletException {
+    private List<RasRunResult> getRunsByIds(List<String> runIds) throws InternalServletException {
         
-		boolean isShort = isTestRunShort(isMethodDetailsRequested);
+		boolean isShort = isTestRunShort(rasDetailsQueryParams.getDetail());
```
eamansour: isShort and isTestRunShort could confuse people - perhaps this should be something like isMethodDetailsExcluded? Or something better?

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -81,9 +82,11 @@ public class RunQueryRoute extends RunsRoute {
 	public static final String QUERY_PARAMETER_CURSOR = "cursor";
 	public static final String QUERY_PARAMETER_RUNNAME = "runname";
 	public static final String QUERY_PARAMETER_RUNID = "runid";
+	public static final String QUERY_PARAMATER_DETAIL = "detail";
```
eamansour: Typo - PARAMATER should be PARAMETER

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
eamansour: Also, should we allow a detail value like`METHODS`,`Methods`, `meThOds`, etc (i.e. mixed case/upper case)?

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -95,7 +95,7 @@ public class RunQueryRoute extends RunsRoute {
 
 
 	private static final GalasaGson gson = new GalasaGson();
-	private final RasDetailsQueryParams rasDetailsQueryParams;
+	private boolean IS_METHOD_DETAILS_EXCLUDED = true;
```
eamansour: I think this field needs removing since excluding method details is done on a per-request basis, we can use a local variable inside the GET method instead

eamansour (approved):

PR #286, opened 2025-05-07
Refactored runs get method

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2220


modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -1586,6 +1586,14 @@ paths:
           schema:
             type: string
           example: a6847035-7768-4416-91a6-fe05cf10da85
+        - name: detail
+          in: query
+          description: |
+            The detail level requested in the request.
+            "methods" value will result in returning method details in a test run
```
eamansour: `detail level` doesn't exactly explain what this query parameter does. Maybe we could say something about it controlling what details are extracted from a run's test structure and included in responses?

Also, is the query parameter case sensitive? i.e. do we allow `mEthOds` or `METHODS`? Would be useful to say if it is/isn't.

modules/cli/pkg/runs/runsQuery.go
```diff
@@ -18,111 +18,119 @@ import (
...
+		}
+		if query.pageCursor != "" {
+			apicall = apicall.Cursor(query.pageCursor)
+		}
+		if query.group != "" {
+			apicall = apicall.Group(query.group)
+		}
+		if query.isNeedingMethodDetails == true {
+			apicall = apicall.Detail(METHOD_DETAIL_QUERY_PARAM)
+		}
```
eamansour: Really minor - we don't need the `== true` when checking true/false values.

modules/cli/pkg/runs/runsGet.go
```diff
@@ -89,27 +89,20 @@ func GetRuns(
 		chosenFormatter, err = validateOutputFormatFlagValue(outputFormatString, validFormatters)
 		if err == nil {
 			var runJson []galasaapi.Run
-			runJson, err = GetRunsFromRestApi(runName, requestorParameter, resultParameter, fromAge, toAge, shouldGetActive, timeService, commsClient, group)
+			isNeedingMethodDetails := chosenFormatter.IsNeedingMethodDetails()
+			runJson, err = GetRunsFromRestApi(runName, requestorParameter, resultParameter, fromAge, toAge, shouldGetActive, timeService, commsClient, group, isNeedingMethodDetails)
```
eamansour: Was expecting some unit test changes as a result of this change - the original code was sending another request to `/ras/runs/{run-id}` for details, but now we only send one request to `/ras/runs` with query parameters (which is great). 

Are we testing that the right query parameters are being sent to the API server now?

modules/cli/pkg/runs/runs.go
```diff
@@ -30,8 +30,9 @@ func getRunIdFromRunName(runName string,
 	fromAgeHours := 0
 	toAgeHours := 0
 	shouldGetActive := true
+	isNeedingMethodDetails := false
 
-	runs, err = GetRunsFromRestApi(runName, requestorParameter, resultParameter, fromAgeHours, toAgeHours, shouldGetActive, timeService, commsClient, group)
+	runs, err = GetRunsFromRestApi(runName, requestorParameter, resultParameter, fromAgeHours, toAgeHours, shouldGetActive, timeService, commsClient, group, isNeedingMethodDetails)
```
eamansour: This list of parameters is getting quite long, maybe we could create a RunQuery object outside and then pass it in to this function?

So for example, we could do something like:
```go
runsQuery := NewRunsQuery(runName, requestor, ...)
runs, err = GetRunsFromRestApi(runsQuery, commsClient)
```

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
eamansour: We could also provide a list of the supported values. In this case, it'll only include "methods".

Then we can say what providing "methods" does, something like:
```
Providing "methods" will include the list of methods that were executed as part of a test run
```

Edit the above as necessary :)

modules/cli/pkg/runs/runsGet.go
```diff
@@ -89,27 +89,23 @@ func GetRuns(
 		chosenFormatter, err = validateOutputFormatFlagValue(outputFormatString, validFormatters)
 		if err == nil {
 			var runJson []galasaapi.Run
-			runJson, err = GetRunsFromRestApi(runName, requestorParameter, resultParameter, fromAge, toAge, shouldGetActive, timeService, commsClient, group)
-			if err == nil {
-				// Some formatters need extra fields filled-in so they can be displayed.
-				if chosenFormatter.IsNeedingMethodDetails() {
-					log.Println("This type of formatter needs extra detail about each run to display")
-					runJson, err = GetRunDetailsFromRasSearchRuns(runJson, commsClient)
```
eamansour, line 97: Is `GetRunDetailsFromRasSearchRuns` still being used elsewhere? If not, might be worth removing it.

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -1586,6 +1586,15 @@ paths:
           schema:
             type: string
           example: a6847035-7768-4416-91a6-fe05cf10da85
+        - name: detail
+          in: query
+          description: |
+            Controls the amount of details extracted from a run's test structure and included in responses.
+            Providing "methods" will include the list of methods that were executed as part of a test run.
+            Only "methods" is supported yet.
```
eamansour: Minor - "yet" isn't needed here.

Still might be worth saying whether the value for the `detail` query parameter is case-sensitive here. Someone could try providing `METHODS` but then be faced with an error?

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/common/RasDetailsQueryParams.java
```diff
@@ -24,4 +26,17 @@ public boolean isMethodDetailsExcluded(String detailParam) {
 
     }
 
+    public boolean isParamSupported(String queryParam) {
+
+        boolean isParamSupported = false;
+
+        for(String type: SUPPORTED_DETAIL_QUERY_PARAMS) {
+            if(type.equals(queryParam))
+            isParamSupported = true;
```
eamansour, line 35: Minor - break out of this loop if we've matched the query parameter value?

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/common/RasDetailsQueryParams.java
```diff
@@ -9,6 +9,8 @@ public class RasDetailsQueryParams {
     
     public static final String DETAILS_METHOD_QUERY_PARAM_VALUE = "methods";
 
+    public static final String[] SUPPORTED_DETAIL_QUERY_PARAMS = {DETAILS_METHOD_QUERY_PARAM_VALUE};
```
eamansour: We have the `SupportedQueryParameterNames` class - maybe we could re-use it somehow here since we're essentially doing the same check?

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/common/RasDetailsQueryParams.java
```diff
@@ -24,4 +26,17 @@ public boolean isMethodDetailsExcluded(String detailParam) {
 
     }
 
+    public boolean isParamSupported(String queryParam) {
```
eamansour, line 31: The SupportedQueryParameterNames class has an `isSupported` method... wonder if we can re-use it here?

eamansour (approved):
Great work :)

PR #292, opened 2025-05-13
Added new endpoint to cancel grouped runs

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2003


modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -2063,6 +2063,102 @@ paths:
...
+  #------------------------------------------
+  /ras/groups/{groupid}:
+    put:
+      summary: Update the status of a test run by group id
+      description: |
+        Updates the status of a test run in order to cancel it.
+
+        To cancel the run, put a value such that the status is "finished" and the result is "cancelled".
+
+        To reset the run, put a value such that the status is "queued" and the result is null or ""
```
jadecarino: Does this story include resetting runs as well as cancelling them? Just wondering should line 2078 be included here.

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -2063,6 +2063,102 @@ paths:
...
+      responses:
+        '202':
+          description: Run status updated
+          content:
+            text/plain:
+              schema:
+                type: string
+              examples:
+                cancelRun:
+                  value: The request to reset run with group id xxgroupxx has been received.
```
jadecarino: This should say `The request to cancel` to match line 2106 above.

jadecarino: Also if this story includes resetting runs, there should be another section for `resetRun`?

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -2063,6 +2063,102 @@ paths:
...
+                invalidStatusUpdateRequest:
+                  value:
+                    error_code: 5045
+                    error_message: "GAL5045E: Error occurred. The field 'status' in the request body is invalid. The 'status' value 'create' supplied is not supported. Supported values are: 'cancel' and 'reset'"
+                  summary: An Error occurred because the status field does not contain an expected action
+                runNameDoesNotMatch:
+                  value:
+                    error_code: 5046
+                    error_message: "GAL5046E: The 'runName' 'badRun' from the request body does not match the  'runName' 'U123'  associated with the runID in the url 'xx12345xx'."
+                  summary: An Error occurred because the runName in the request body does not match the runName derived from runID in the URL
```
jadecarino: Would this be valid here? As there wouldn't be a `runName` being provided in the request body here. 

Is the behaviour that it loops through each run in a group and has multiple responses for each run?

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -2063,6 +2063,102 @@ paths:
...
+                resetRunHasAlreadyCompleted:
+                  value:
+                    error_code: 5049
+                    error_message: "GAL5049E: Error occurred when trying to reset the run ''{0}''. The run has already completed."
+                  summary: The run in the request has already completed and can not be reset
+                cancelRunHasAlreadyCompleted:
+                  value:
+                    error_code: 5050
+                    error_message: "GAL5050E: Error occurred when trying to cancel the run 'U123'. The run has already completed."
+                  summary: The run in the request has already completed and can not be cancelled
```
jadecarino: Similar comment here - these responses are about individual runs, will this API call respond with multiple responses about each run in a group (what these lines seem like)?

Or would it have one combined response at the end after trying to cancel each run in the group?

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/GroupDetailsRoute.java
```diff
@@ -0,0 +1,100 @@
...
+
+      String responseBody = "";
+      List<IRunResult> groupedRuns = getRunsByGroupId(groupId);
+
+      if (groupedRuns != null && !groupedRuns.isEmpty()) {
+
+         for (IRunResult run : groupedRuns) {
+
+            RunActionStatus status = RunActionStatus.getfromString(runAction.getStatus());
+            String result = runAction.getResult();
```
jadecarino: Perhaps these two lines should be outside of the loop as it looks like they will be the same each iteration as their values come from the `runAction` which is passed in

jadecarino: Just to avoid running these lines multiple times if the value will always be the same.

jadecarino: If there is a reason they are in the loop that I am missing let me know :)

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/GroupDetailsRoute.java
```diff
@@ -0,0 +1,100 @@
...
+               ServletError error = new ServletError(GAL5045_INVALID_STATUS_UPDATE_REQUEST, runAction.getStatus());
+               throw new InternalServletException(error, HttpServletResponse.SC_BAD_REQUEST);
+            } else if (status == RunActionStatus.QUEUED) {
+               runStatusUpdate.resetRun(runName);
+               logger.info("Run reset by external source.");
+               responseBody = String.format("The request to reset run with group id %s has been received.", groupId);
+            } else if (status == RunActionStatus.FINISHED) {
+               runStatusUpdate.cancelRun(runName, result);
+               logger.info("Run cancelled by external source.");
+               responseBody = String.format("The request to cancel run with group id %s has been received.", groupId);
```
jadecarino: If a group has multiple runs, then this `responseBody` is going to be overwritten multiple times, with the only one being returned at the end of the loop being for the last run in the array. Is this the intended behaviour?

Perhaps instead you could concatenate the strings for the multiple runs, store that in an overall `responseBody` and return that, to ensure you have output for each run in the group.

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/test/java/dev/galasa/framework/api/ras/internal/routes/TestGroupDetailsRoute.java
```diff
@@ -0,0 +1,559 @@
...
+import dev.galasa.framework.spi.IFrameworkRuns;
+import dev.galasa.framework.spi.IResultArchiveStoreDirectoryService;
+import dev.galasa.framework.spi.IRun;
+import dev.galasa.framework.spi.IRunResult;
+
+import static org.assertj.core.api.Assertions.*;
+
+public class TestGroupDetailsRoute extends RasServletTest {
+
+	Pattern grouPattern = Pattern.compile(GroupDetailsRoute.path);
```
jadecarino: Minor comment but a typo here

jadecarino (changes requested):
Hi Ash, good work, I have a few comments throughout, but they are all related to one of two questions I have about the work item itself, that perhaps if I understand better, will address my comment.

1) Is this work item meant to include resetting runs by their groupId or just cancelling them? The story only mentions cancelling them, but if you have had a separate discussion about including resetting also, thats fine.

2) My main question - how is the processing of the request and returning of a response meant to work? A groupId is provided to the API, which finds a list of matching runs. Then that list is looped through and each individual run is attempted to cancel. Are we trying to send multiple responses back to the client for if each run was successfully cancelled/not cancelled due to error, or are we meant to be combining this into a "overall" response and sending that back? This will help me understand better whats trying to be achived.

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -2140,6 +2140,82 @@ paths:
...
+      description: |
+        Updates the status of a test run in order to cancel it.
+
+        To cancel the run, put a value such that the status is "finished" and the result is "cancelled".
+
+        Requests to this endpoint require a valid bearer token in JWT format to be provided
+        in the 'Authorization' header (e.g. 'Authorization: Bearer <bearer-token>').
+      operationId: putRasRunStatusByGroupId
+      tags:
+      - Result Archive Store API
```
jadecarino: This should be `Runs API` now that it's moved path I believe

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -2934,6 +3010,12 @@ components:
         result:
           type: string
           enum: [nil,"cancelled"]
+    GroupCancelRunStatusRequest:
```
jadecarino: You haven't referenced this anywhere, should this be on line 2167?

Also, this should have a `status` section I believe like the one above? If we are just concerned with cancelling runs at the moment, then I believe the only valid result would be "cancelled" and status would be "finished".

modules/framework/galasa-parent/dev.galasa.framework.api.runs/src/main/java/dev/galasa/framework/api/runs/common/GroupRunActionJson.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package dev.galasa.framework.api.runs.common;
+
+public class GroupRunActionJson {
+
+    private String result;
+
+    public GroupRunActionJson(String status, String result){
```
jadecarino: Do we need `status` here if its not being used? Why is it not being used?

modules/framework/galasa-parent/dev.galasa.framework.api.runs/src/test/java/dev/galasa/framework/api/runs/routes/TestGroupRunsRoute.java
```diff
@@ -735,4 +735,121 @@ public void testPostUUIDGroupNameRunsWithValidBodyAndMultipleClassesReturnsWithR
...
+        String payload = generateStatusUpdateJson("cancelled");
+        String submissionId = "submission1";
+        Set<String> tags = new HashSet<>();
+        addRun("name1", "type1", "requestor1", "test1", "BUILDING","bundle1", "testClass1", groupName, submissionId,tags);
+        addRun("name2", "type2", "requestor2", "test2", "BUILDING","bundle2", "testClass2", groupName, submissionId,tags);
+        addRun("name3", "type3", "requestor3", "test3", "BUILDING","bundle3", "testClass3", groupName, submissionId,tags);
+        addRun("name4", "type4", "requestor4", "test4", "BUILDING","bundle4", "testClass4", groupName, submissionId,tags);
+        addRun("name5", "type6", "requestor5", "test5", "BUILDING","bundle5", "testClass6", groupName, submissionId,tags);
+        addRun("name6", "type6", "requestor6", "test6", "FINISHED","bundle6", "testClass6", groupName, submissionId,tags);
+        addRun("name7", "type7", "requestor7", "test7", "FINISHED","bundle7", "testClass7", groupName, submissionId,tags);
```
jadecarino: This test case says all the runs in the group should be active so should these two not be `FINISHED`?

modules/framework/galasa-parent/dev.galasa.framework.api.runs/src/main/java/dev/galasa/framework/api/runs/common/GroupRunActionStatus.java
```diff
@@ -0,0 +1,33 @@
...
+public enum GroupRunActionStatus {
+
+    CANCELLED("cancelled"),
+    ;
+    private String value;
+
+    private GroupRunActionStatus(String type){
+        this.value = type;
+    }
+
```
jadecarino, line 17: I'm not sure I understand what's happening here

modules/framework/galasa-parent/dev.galasa.framework.api.runs/src/main/java/dev/galasa/framework/api/runs/routes/GroupRunsRoute.java
```diff
@@ -83,4 +89,68 @@ public HttpServletResponse handlePostRequest(
...
+
+        return getResponseBuilder().buildResponse(request, res, "text/plain", responseBody, responseStatusCode);
+    }
+
+    private String updateRunStatus(String groupId, GroupRunActionJson runAction)
+            throws InternalServletException, ResultArchiveStoreException {
+
+        String responseBody = "";
+        List<IRun> groupedRuns = getRuns(groupId);
+        GroupRunActionStatus cancelAction = GroupRunActionStatus.getfromString(runAction.getResult());
```
jadecarino: I think I'm finding it hard to follow why we are essentially storing the result of "cancelled" in two different things - the `GroupRunActionJson` then it seems like we just transfer it to the `GroupRunActionStatus` - can you explain please why we need both?

jadecarino (changes requested):
Hi Ash, just a few small changes, and also a question perhaps just to help my understanding of whats going on. Thanks!

jadecarino (approved):
Looks good! Thanks for making these changes Ash

PR #294, opened 2025-05-15
Improve Advanced CodeQL workflows

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2228

Due to the complexity of this repo's structure, Default CodeQL cannot find the Galasa CLI code to scan it. Additionally, due to the fact that CodeQL only knows to look for dependencies in Maven Central, it can't find the `dev.galasa.platform` bundle and the Maven/Gradle plugins bundles, so therefore can't form a full dependency tree (meaning the scan isn't comprehensive).

To get around this, you must write an Advanced CodeQL workflow (essentially a workflow that calls `codeql-action/init` and `codeql-action/analyze`) that first builds the source code so:
1) Java - its available in the local Maven repo of the GitHub Actions runner (`/home/runner/.m2/repository`), that way CodeQL can find it.
2) Go - when the `make` command is invoked, CodeQL monitors the build command in real time to determine what code to analyze.

This PR updates the existing Github workflows in this repo to support the above.

1) At the end of the Pull Request Build Orchestrator, as the code has been built or downloaded as part of the PR build, it calls the CodeQL Java and/or Go scans based on if the Java modules and/or Go modules have changed.
2) At the end of the Main Build Orchestrator, as the code has been built, it calls the CodeQL Java and Go scans.
3) CodeQL Java and Go scans scheduled weekly.

I have had to make some additions to two gradle.properties files as part of this - some build.gradle files sometimes use parameters that are usually passed in for `$sourceMaven` etc, but as CodeQL's scanning doesn't pass anything in, the build partially fails as it doesn't know the value. The gradle.properties provides a default value.


.github/workflows/codeql-go.yml
```diff
@@ -54,6 +50,38 @@ jobs:
...
+    - name: Download All 'dev.galasa' Artifacts
+      if: ${{ github.event_name == 'pull_request' || github.event_name == 'push' }}
+      uses: actions/download-artifact@v4
+      with:
+        name: all-artifacts
+        path: /home/runner/.m2/repository/dev/galasa
+
+    # If this is running on a schedule, we need to
+    # download from the last Main Build Orchestrator
+    - name: Find Artifacts
```
eamansour: Really minor but for clarity, we could rename this step to something like "Find artifacts from the last successful main build workflow"? Or "Get last successful main build workflow ID"?

Same applies for the Java CodeQL workflow

eamansour (approved):
One minor comment but looks good!

.github/workflows/codeql-go.yml
jadecarino: Good point - I have updated the step names in both workflows.

PR #298, opened 2025-05-21
CLI now supports cancelling runs by group

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2003

- **Added group cancel functionality to CLI**
- **Updated openapi  responses**


modules/cli/docs/generated/errors-list.md
```diff
@@ -243,10 +243,15 @@ The `galasactl` tool can generate the following errors:
 - GAL1244E: Failed to delete stream {}. Unexpected http status code {} received from the server. Error details from the server are not in a valid json format. Cause: '{}'
 - GAL1245E: Failed to delete stream {}. Unexpected http status code {} received from the server. Error details from the server are: '{}'
 - GAL1246E: Failed to delete stream {}. Unexpected http status code {} received from the server. Error details from the server are not in the json format.
+- GAL1247E: Error cancelling runs with group name '{}. Reason: '{}'
```
eamansour: Minor - the `{}` for group name is missing a closing `'`

modules/cli/docs/generated/errors-list.md
```diff
@@ -243,10 +243,15 @@ The `galasactl` tool can generate the following errors:
...
 - GAL2000W: Warning: Maven configuration file settings.xml should contain a reference to a Galasa repository so that the galasa OBR can be resolved. The official release repository is '{}', and 'pre-release' repository is '{}'
 - GAL2501I: Downloaded {} artifacts to folder '{}'
 
 - GAL2503I: The request to reset run '{}' has been accepted by the server.
 
 - GAL2504I: The request to cancel run '{}' has been accepted by the server.
 
+- GAL2505I: The request to cancel runs with group id '{} has been accepted by the server.
```
eamansour: Minor - this mentions "group id", should this be "group name"? Also the `{}` is missing a closing `'`

modules/cli/docs/generated/errors-list.md
```diff
@@ -243,10 +243,15 @@ The `galasactl` tool can generate the following errors:
...
+
+- GAL2506I: The request to cancel runs with group id '{} has been handled successfully. However, no recent active (unfinished) test runs were found which are part of that group. Archived test runs may be part of that group, which can be queried separately from the Result Archive Store..
```
eamansour: Same here regarding group id vs group name and the missing closing `'`

Also, the end has an extra `.`

modules/cli/docs/generated/galasactl_runs_cancel.md
```diff
@@ -13,8 +13,9 @@ galasactl runs cancel [flags]
 ### Options
 
 ```
-  -h, --help          Displays the options for the 'runs cancel' command.
-      --name string   the name of the test run to cancel
+      --group string   An optional flag indicating the name of the test group to cancel. Cannot be used with --name
+  -h, --help           Displays the options for the 'runs cancel' command.
+      --name string    An optional flag indicating the name of the test run to cancel. Cannot be used with --group
```
eamansour: Both `--group` and `--name` are said to be optional, but we need at least one to be passed in. Maybe just remove the mention of `optional` here?

modules/cli/pkg/cmd/runsCancel.go
```diff
@@ -84,9 +85,11 @@ func (cmd *RunsCancelCommand) createRunsCancelCobraCmd(factory spi.Factory,
 		},
 	}
 
-	runsCancelCmd.PersistentFlags().StringVar(&cmd.values.runName, "name", "", "the name of the test run to cancel")
+	runsCancelCmd.PersistentFlags().StringVar(&cmd.values.runName, "name", "", "An optional flag indicating the name of the test run to cancel. Cannot be used with --group")
+	runsCancelCmd.PersistentFlags().StringVar(&cmd.values.group, "group", "", "An optional flag indicating the name of the test group to cancel. Cannot be used with --name")
 
-	runsCancelCmd.MarkPersistentFlagRequired("name")
+	runsCancelCmd.MarkFlagsMutuallyExclusive("group", "name") // Only one of these flags can be used at a time.
+	runsCancelCmd.MarkFlagsOneRequired("name", "group")       // At least one of these flags must be used.
```
eamansour: Really minor but these inline comments aren't really necessary since the MarkFlagsMutuallyExclusive and MarkFlagsOneRequired are descriptive enough in my opinion

modules/cli/pkg/runs/runsCancel.go
```diff
@@ -29,36 +29,49 @@ func CancelRun(
...
-	if runName == "" {
-		err = galasaErrors.NewGalasaError(galasaErrors.GALASA_ERROR_MISSING_NAME_FLAG, runName)
-	}
-
 	if (err == nil) && (runName != "") {
 		err = ValidateRunName(runName)
 	}
 
+	if err == nil && (group != "") {
+		group, err = validateGroupname(group)
```
eamansour: Minor - can this function use camelCase like `validateGroupName` instead of groupname

modules/cli/pkg/runs/runs.go
```diff
@@ -16,6 +16,8 @@ import (
 	"github.com/galasa-dev/cli/pkg/spi"
 )
 
+const GROUP_CANELLED_RESULT = "cancelled"
```
eamansour: Typo - CANELLED

modules/cli/pkg/runs/runsCancel.go
eamansour: `err` is always going to be nil at the start since you've moved the name flag check up to be in the cobra command, the `if` conditions can be simplified by removing the `err == nil` bits.

modules/cli/pkg/runs/runsCancel.go
```diff
@@ -29,36 +29,49 @@ func CancelRun(
...
+	}
+
 	if err == nil {
 
-		runId, err = getRunIdFromRunName(runName, timeService, commsClient)
+		if runName != "" {
```
eamansour: This can be simplified by moving the cancel by run name up after the validateRunName call, likewise for the group name processing.

Perhaps they could be in two smaller functions as well - one for cancelling a run by run name and another for cancelling runs by group?

modules/cli/pkg/runs/runsCancel.go
```diff
@@ -119,3 +131,74 @@ func cancelRun(runName string,
...
+
+func writeConsoleMessage(console spi.Console, errorMessage galasaErrors.MessageType, groupName string) error {
+	var err error
+
+	consoleErr := console.WriteString(fmt.Sprintf(errorMessage.Template, groupName))
+
+	if consoleErr != nil {
+		err = consoleErr
+	}
+	return err
```
eamansour: I don't think `consoleErr` is needed here? `Seems like this function isn't doing much aside from the console.WriteString call?

This method is equivalent to:
```
return console.WriteString(...)
```

modules/cli/pkg/runs/runsCancel_test.go
```diff
@@ -47,6 +47,29 @@ func WriteMockRasRunsPutStatusFinishedResponse(
...
+	} else if groupName == "finished-runs-group" {
+		statusCode = 200
+		response = fmt.Sprintf("Info: When trying to cancel the run group '%s', no recent active (unfinished) test runs were found which are part of that group. Archived test runs may be part of that group, which can be queried separately from the Result Archive Store.", groupName)
+		writer.Header().Set("Content-Type", "text/plain")
+	}
+
+	writer.WriteHeader(statusCode)
+	writer.Write([]byte(response))
+}
+
```
eamansour: We could replace this with individual HTTP interactions in the unit tests, that way each unit test will clearly show what responses are being returned from the mock server.

modules/cli/pkg/runs/runsCancel_test.go
```diff
@@ -70,6 +93,25 @@ func NewRunsCancelServletMock(
...
+		acceptHeader := req.Header.Get("Accept")
+
+		assert.Equal(t, "application/json", acceptHeader, "Expected Accept: application/json header, got: %s", acceptHeader)
+		WriteMockGroupRunsPutStatusFinishedResponse(t, writer, req, groupName)
+
+	}))
+
+	return server
+}
+
```
eamansour: Same here regarding HTTP interactions - see the `runs get` unit tests for an example of this.

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -2149,7 +2149,7 @@ paths:
 
         Requests to this endpoint require a valid bearer token in JWT format to be provided
         in the 'Authorization' header (e.g. 'Authorization: Bearer <bearer-token>').
-      operationId: putRasRunStatusByGroupId
+      operationId: putRunStatusByGroupId
```
eamansour, line 2152: This seems like a small change but does this impact OpenAPI-generated code? If we change the operationId, then any generated functions that use this will change, which could impact users that generate code from our OpenAPI docs.

modules/cli/pkg/runs/runs.go
eamansour: Also, the OpenAPI docs define the cancelled result as an enum, are we not able to use that value directly from the OpenAPI-generated code?

modules/framework/galasa-parent/dev.galasa.framework.api.runs/src/test/java/dev/galasa/framework/api/runs/RunsServletTest.java
```diff
@@ -111,7 +110,7 @@ protected String generateExpectedString(String groupId, int statusCode) {
         if(statusCode == 202) {
             responseMessage = String.format("The request to cancel run with group id %s has been received.", groupId);
         } else if (statusCode == 200){
-            responseMessage = String.format(GAL5430_GROUP_RUNS_ALREADY_FINISHED.toString(), groupId);
+            responseMessage = String.format("Info: When trying to cancel the run group '%s', no recent active (unfinished) test runs were found which are part of that group. Archived test runs may be part of that group, which can be queried separately from the Result Archive Store.", groupId);
```
eamansour: If we're only checking a string of text, we could put this directly in the tests that use it? That means we won't need the if statements and it would be clearer what the tests actually expect since the assertions will be closer to the tests themselves.

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
aashir21: > This seems like a small change but does this impact OpenAPI-generated code? If we change the operationId, then any generated functions that use this will change, which could impact users that generate code from our OpenAPI docs.

Just being used in one place, that too with what im working with currently. So updated the code in CLI as well. No issues I think

modules/cli/pkg/runs/runsCancel_test.go
```diff
@@ -270,19 +228,18 @@ func TestRunsCancelWhereServerSideResponseCannotBeParsedReturnsError(t *testing.
 	assert.ErrorContains(t, err, "GAL1136")
 }
 
-func TestGroupRunsCancelWithInvalidGroupNameReturnsError(t *testing.T) {
+func skipTestGroupRunsCancelWithInvalidGroupNameReturnsError(t *testing.T) {
```
eamansour: This test is being skipped?

modules/cli/pkg/errors/errorMessage.go
```diff
@@ -464,4 +465,7 @@ var (
 	GALASA_INFO_FOLDER_DOWNLOADED_TO = NewMessageType("GAL2501I: Downloaded %d artifacts to folder '%s'\n", 2501, STACK_TRACE_NOT_WANTED)
 	GALASA_INFO_RUNS_RESET_SUCCESS   = NewMessageType("GAL2503I: The request to reset run '%s' has been accepted by the server.\n", 2503, STACK_TRACE_NOT_WANTED)
 	GALASA_INFO_RUNS_CANCEL_SUCCESS  = NewMessageType("GAL2504I: The request to cancel run '%s' has been accepted by the server.\n", 2504, STACK_TRACE_NOT_WANTED)
+
+	GALASA_INFO_GROUP_RUNS_CANCEL_SUCCESS   = NewMessageType("GAL2505I: The request to cancel runs with group name '%s' has been accepted by the server.\n", 2505, STACK_TRACE_NOT_WANTED)
+	GALASA_INFO_GROUP_RUNS_ALREADY_FINISHED = NewMessageType("GAL2506I: The request to cancel runs with group name '%s' has been handled successfully. However, no recent active (unfinished) test runs were found which are part of that group. Archived test runs may be part of that group, which can be queried separately from the Result Archive Store..\n", 2506, STACK_TRACE_NOT_WANTED)
```
eamansour, line 470: Are these being used? Might be better than the hardcoded strings in the routes? Still has the extra '.' at the end of 2506I.

eamansour: Still has the trailing `.` after `Result Archive Store..`

eamansour (approved):

PR #301, opened 2025-05-21
3270 manager: set initial cursor position and send up to 256 bytes of outbound data for SSCP-LU session screens

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2232

SSCP-LU session displays are unformatted and use an "initial cursor position" to determine the start of an input area. The initial cursor position should be set to one screen position after the last character in the SSCP-LU inbound data stream, or the start of the next line if the last character was a newline.

The input area has an outbound limit of 256 bytes of data, data entry is restricted from the initial cursor position to the following 255 screen positions or to the last screen position.

This PR updates the 3270 manager to take these attributes of SSCP-LU session displays into account when entering data into such displays.

## Changes
- An initial cursor position is now stored after processing inbound messages
- When building the outbound message bytes, SSCP-LU session screens will build messages starting from the initial cursor position onwards as described above
- Added an insert cursor instruction when converting SSCP-LU data streams into standard 3270 text, which moves the cursor to one position after the last character/the start of the next line in the event of the last character being a newline
  - Note: The 3270 manager already had code that processed unformatted screens in standard 3270 sessions (i.e. not SSCP-LU sessions). This existing processing has been left as-is, since those screens are intended to be sent in their entirety


modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/internal/comms/NetworkThread.java
```diff
@@ -217,11 +217,13 @@ public void processMessage(InputStream messageStream) throws IOException, Networ
             // This advances through the message buffer to skip the header bytes after DATA-TYPE.
             buffer.get(new byte[4]);
             
+            boolean isSSCPLUDisplay = false;
```
techcobweb: This whole method is getting pretty big. Is there any way to split it up into su-functions to make it more readable ?

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/spi/Screen.java
```diff
@@ -104,6 +104,9 @@ public class Screen {
     private final int                               alternateColumns;
     private final int                               alternateRows;
 
+    private boolean                                 isSSCPLUDisplay = false;
```
techcobweb, line 111: Some comment as to what an SSCPLU display actually is ?

techcobweb (approved):

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/internal/comms/NetworkThread.java
eamansour: Addressed in https://github.com/galasa-dev/galasa/pull/301/commits/465cb1069e44853227cb9821fc91c43e49bb66fc

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/spi/Screen.java
eamansour: Done in https://github.com/galasa-dev/galasa/pull/301/commits/465cb1069e44853227cb9821fc91c43e49bb66fc

PR #302, opened 2025-05-22
Fix NullPointerException when interrupting runs that have no RAS record

eamansour:
## Why?
Likely related to https://github.com/galasa-dev/projectmanagement/issues/2231

Runs that are in the `allocated` state have no RAS record yet, and when these runs fail to start after a given period of time, the dead heartbeat monitor times them out and tries to mark them as "Hung", which causes a null pointer exception as the engine controller attempts to update the runs' non-existent RAS records.

This PR adds a guard against null run IDs when setting and processing a run's RAS actions.

## Changes
- Guard against null run IDs when processing a run's RAS actions
- When a run is marked as interrupted and the run has no RAS run ID, no RAS actions will be set for the run in the DSS - only an interrupt reason will be set


jadecarino (approved):
Looks good!

PR #306, opened 2025-05-23
Remove use of internal bootstrap for test pods and remove internal bootstrap endpoint

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2233

## Changes
- Added a `GALASA_EXTRA_BUNDLES` environment variable that can be used to set the `framework.extra.bundles` bootstrap property
- Test pods no longer use the API server's internal bootstrap when launched, all bootstrap properties are now loaded from environment variables attached to the test pod's container
- Replaced the internal bootstrap endpoint with the external bootstrap endpoint (i.e. moved `/bootstrap/external` to `/bootstrap`) - this will have no impact on client tools since `/bootstrap/external` was never published on our OpenAPI docs and the `/bootstrap` endpoint in existing Galasa services serves up the external bootstrap


techcobweb:
I think we need to add something to the readme about this, perhaps the readme in the framework ? We don't currently document the `java -jar` syntax, but it seems that we should ? I know this is 'internals' but it might help if we have somewhere to document this stuff, as the galasactl client uses it, and other clients could use it also.

techcobweb (approved):

modules/framework/galasa-parent/galasa-boot/src/main/java/dev/galasa/boot/Launcher.java
```diff
@@ -289,8 +291,14 @@ private void processCommandLine(String[] args) throws ParseException {
             }
         }
 
-        checkForBootstrap(commandLine);
+        if (!commandLine.hasOption(NO_BOOTSTRAP_OPTION)) {
```
eamansour: No need, change the default behaviour of `checkBootstrap` to initialise a blank Properties object.

eamansour: Done in https://github.com/galasa-dev/galasa/pull/306/commits/7cf616c9bb0f8830b3226cf0b3e5edae1a41367a

PR #320, opened 2025-06-02
build-locally scripts changed to use local .m2 rather than the development site

techcobweb:

# Why ?
- The https site development.galasa.dev is down. 
- Local builds of the mono should not rely on that site being up for them to work

- [x] Changed various build-locally.sh scripts to remove this dependency.
- [x] Stopped the CLI build-locally.sh running tests, as to get those working again needs an actual change to the CLI build `--development` flag, which isn't needed just to get things working again. We can re-instate the tests at a later point.

### DIFF
--- a/modules/cli/build-locally.sh
+++ b/modules/cli/build-locally.sh
@@ -59,6 +59,11 @@ Options are:
 -d | --delta : Do a delta build. One of the --clean or --delta flags are mandatory.
 -s | --detectsecrets true|false : Do we want to detect secrets in the entire repo codebase ? Default is 'true'. Valid values are 'true' or 'false'
 
+Environment Variables:
+SOURCE_MAVEN :
+    Used to indicate where parts of the OBR can be obtained.
+    Optional. Could be set to something like: https://development.galasa.dev/main/maven-repo/obr/
+    Defaults to file://users/xxx/.m2/repository
 EOF
 }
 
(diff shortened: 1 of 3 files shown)


techcobweb:
Planning to force-merge these changes so people are un-blocked locally.

modules/cli/build-locally.sh
```diff
@@ -589,7 +606,7 @@ function run_test_locally_using_galasactl {
     --obr mvn:${OBR_GROUP_ID}/${OBR_ARTIFACT_ID}/${OBR_VERSION}/obr \
     --class ${BUNDLE}/${JAVA_CLASS} \
     --class ${BUNDLE}/${JAVA_CLASS_2} \
-    --remoteMaven ${REMOTE_MAVEN} \
+    --remoteMaven ${SOURCE_MAVEN} \
```
jadecarino, line 609: Should this flag be `--localMaven` instead if the default value of `SOURCE_MAVEN` is likely to be a local directory now not a remote url?

jadecarino: No strong preference here, as it obviously could be overridden back to a remote, just thought I'd mention

jadecarino (approved):
Approved with just one small comment for the CLI script but no issue with you merging this as is. You should be able to merge without force merging - the GitHub Actions builds should still pass, they download the artifacts from other workflows, not the download site.

modules/cli/build-locally.sh
techcobweb: I turned off the tests, so it's kind of a moot point until we get them re-enabled.

The tests have a problem in that they use galasactl local init --development, which burns-in the dependency on development.galasa.dev. So had to just not call them so far.

Will return to this when people are able to build the mono repo locally I think.

PR #330, opened 2025-06-05
Add Docker-based Isolated and MVP compilation tests

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2190

## Changes
- Added a new `dev.galasa.ivts.compilation` bundle with tests that use the Docker manager to run a `gradle build` command against an example test project that exists inside the new Docker images pulled in the tests
- Added a Dockerfile that downloads a build of Galasa (parameterised via a given URL), creates an example project with `galasactl project create`, and copies over a `build.gradle` file that depends on the test being run
  - Two `build.gradle` files have been checked in, one for the isolated test and one for the MVP test, so one Docker image can be created per Galasa build type
  - The dependencies listed in these build.gradle files have been taken from the existing CompilationLocal inttests
- Updated the IVTs workflow to include a job that builds the new Docker images for the compilation tests to pull


jadecarino (approved):
Looks great Eamonn. Thanks for putting this together, really excited we've got the last batch of those old inttest converted now.

PR #338, opened 2025-06-10
Add CPS property for users to optionally supply a custom 3270 terminal device name

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2188

## Changes
- Added a new `zos3270.image.IMAGEID.device.name` CPS property to allow users to supply a custom device name during the 3270 terminal negotiation process
- Updated the terminal negotiation code to add the `CONNECT <device-name>` option into the `DEVICE-TYPE REQUEST` command if the new CPS property is set. See sections 4 and 7 in https://www.rfc-editor.org/rfc/rfc2355 for reference.
- Updated the 3270 manager docs to include the new CPS property


modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/internal/properties/TerminalDeviceName.java
```diff
@@ -0,0 +1,30 @@
...
+ * The property takes the following format:
+ * zos3270.image.IMAGEID.device.name=LU1
+ *
+ * If no device name is given, a null value will be returned.
+ */
+public class TerminalDeviceName extends CpsProperties {
+
+    public static String get(IZosImage image) throws Zos3270ManagerException {
+        try {
+            return getStringNulled(Zos3270PropertiesSingleton.cps(), "image", "device.name", image.getImageID());
```
techcobweb: Is there a max length we should enforce ? What are the valid characters for a device name ? Both limitations should be documented in the parameter.

techcobweb (dismissed):
Think we need some validation on the device name ?

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/internal/properties/TerminalDeviceName.java
eamansour: Added validation based on the RFC details https://www.rfc-editor.org/rfc/rfc2355 and updated CPS property docs

jadecarino (approved):
Looks good to me!

PR #347, opened 2025-06-18
k8s controller waits for etcd and RAS to be ready before scheduling new runs

eamansour:
## Why?
See https://github.com/galasa-dev/projectmanagement/issues/2235

## Changes
- The k8s controller now checks that the etcd and RAS pods are in the "Ready" state before proceeding to schedule test runs
- Passed in the Galasa installation name provided when installing the Helm chart into the k8s controller via a `GALASA_INSTALL_NAME` environment variable, so that the k8s controller can query for pods with the `app=<name>-etcd` and `app=<name>-ras` labels
- Removed unused imports


jadecarino (approved):
All looks good to me

PR #348, opened 2025-06-18
Allow spaces in JVM launch options

ideeley:
Implementation of https://github.com/galasa-dev/projectmanagement/issues/2279

### DIFF
--- a/modules/cli/pkg/embedded/templates/galasahome/bootstrap.properties
+++ b/modules/cli/pkg/embedded/templates/galasahome/bootstrap.properties
@@ -15,11 +15,18 @@
 #   If the file does not exist already, then the Galasa framework will try to create any empty file.
 #
 # galasactl.jvm.local.launch.options - Add a space-separated string which contains all the extra options
-#   you want added to a command which launches the JVM to run a test in that local JVM.
+#   you want added to a command which launches the JVM to run a test in that local JVM. If a desired option
+#   contains spaces, this can be specified by escaping spaces with a \ character, or by enclosing them in
+#   unescaped quotes. If an option contains the escape character, it MUST be escaped. Unescaped quotes are 
+#   removed and escaped characters are unescaped as part of parameter processing.
 #   For example:
 #   galasactl.jvm.local.launch.options=-Xmx80m
 #   or
 #   galasactl.jvm.local.launch.options=-Xmx80m -Xms20m
+#   or
+#   galasactl.jvm.local.launch.options=-Djavax.ssl.net.trustStore="C:\\Program Files\\security\\truststore"
+#   or
+#   galasactl.jvm.local.launch.options=-Djavax.ssl.net.trustStore=C:\\Program\ Files\\security\\truststore"
 #
 # galasactl.jvm.local.launch.debug.mode - Only used when the --debug flag is used on launching a test 
 #   within a local JVM. 
--- a/modules/cli/pkg/embedded/templates/galasahome/bootstrap.properties
+++ b/modules/cli/pkg/embedded/templates/galasahome/bootstrap.properties
@@ -16,17 +16,15 @@
 #
 # galasactl.jvm.local.launch.options - Add a space-separated string which contains all the extra options
 #   you want added to a command which launches the JVM to run a test in that local JVM. If a desired option
-#   contains spaces, this can be specified by escaping spaces with a \ character, or by enclosing them in
-#   unescaped quotes. If an option contains the escape character, it MUST be escaped. Unescaped quotes are 
-#   removed and escaped characters are unescaped as part of parameter processing.
+#   contains spaces, this can be specified by enclosing them in double quotes. If a desired option contains
+#   double quotes, this can be specified by escaping the double quotes with a backslash (\) character. Enclosing
+#   double quotes are removed and escaped double quotes are unescaped as part of parameter processing.
 #   For example:
 #   galasactl.jvm.local.launch.options=-Xmx80m
 #   or
 #   galasactl.jvm.local.launch.options=-Xmx80m -Xms20m
 #   or
-#   galasactl.jvm.local.launch.options=-Djavax.ssl.net.trustStore="C:\\Program Files\\security\\truststore"
-#   or
-#   galasactl.jvm.local.launch.options=-Djavax.ssl.net.trustStore=C:\\Program\ Files\\security\\truststore"
+#   galasactl.jvm.local.launch.options=-Dmade.up.property="Something with spaces and \"double quotes\"" -Xmx80m
 #
 # galasactl.jvm.local.launch.debug.mode - Only used when the --debug flag is used on launching a test 
 #   within a local JVM. 
--- a/modules/cli/pkg/launcher/jvmLauncher.go
+++ b/modules/cli/pkg/launcher/jvmLauncher.go
@@ -792,36 +792,28 @@ func appendArgsBootstrapJvmLaunchOptions(args []string, bootstrapProperties prop
 		// Process each character in turn
 		var argBuilder strings.Builder
 		for i, inQuotes := 0, false; i < len(launchOptionChars); i++ {
-			if !inQuotes && (launchOptionChars[i] == api.BOOTSTRAP_PROPERTY_NAME_LOCAL_JVM_LAUNCH_OPTIONS_SEPARATOR) {
-				// If we've reached an unquoted space, that marks the end of the argument so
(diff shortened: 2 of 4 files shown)


modules/cli/pkg/embedded/templates/galasahome/bootstrap.properties
```diff
@@ -15,11 +15,18 @@
...
+#   unescaped quotes. If an option contains the escape character, it MUST be escaped. Unescaped quotes are 
+#   removed and escaped characters are unescaped as part of parameter processing.
 #   For example:
 #   galasactl.jvm.local.launch.options=-Xmx80m
 #   or
 #   galasactl.jvm.local.launch.options=-Xmx80m -Xms20m
+#   or
+#   galasactl.jvm.local.launch.options=-Djavax.ssl.net.trustStore="C:\\Program Files\\security\\truststore"
+#   or
+#   galasactl.jvm.local.launch.options=-Djavax.ssl.net.trustStore=C:\\Program\ Files\\security\\truststore"
```
techcobweb: Does that mean someone already using Galasa with a settings like this will break ?
```
Djavax.ssl.net.trustStore=C:\myFolder\andMyfile
```
? 
What will it do when the `\m`and teh `\a` get reached ?

ideeley: That's correct. That line will be passed to the JVM as `Djavax.ssl.net.trustStore=C:myFolderandMyfile`

ideeley: A safer option might be to only allow escaping of double quote characters, just in case someone wants to pass an option that contains quotes. Also avoids the gnarly issue of escaped characters that traditionally have a special meaning e.g. \n, \r, etc. This would still break anyone currently passing `\"` but I think that's somewhat unlikely.

That would mean that the only way to pass my example option would be `-Djavax.ssl.net.trustStore="C:\Program Files\security\truststore"`, which would be passed as `-Djavax.ssl.net.trustStore=C:\Program Files\security\truststore`.

And if someone wanted to pass a double quote character to the JVM then they could do something like `-DbackslashAndQuote=\\"` which would pass `-DbackslashAndQuote=\"`.

Alternatively, we could support passing quotes by having a double double-quote within double-quotes (as supported by some programming languages) e.g. the example above would be something like `-DbackslashAndQuote="\"""` to achieve the same result.

Let me know what you'd prefer and I can rework it.

techcobweb: I think I prefer this behaviour:
```
-Djavax.ssl.net.trustStore="C:\Program Files\security\truststore"
```
... which would be passed as 
```
arg[0]: -Djavax.ssl.net.trustStore
arg[1]: C:\Program Files\security\truststore
```

So \" inside the quotes would need to be respected as an escaped double-quote I guess.

That's more along the lines of what a bash script does. I was surprised that the standard java properties reader strips off quotes in the first place.

Thoughts on that ?

ideeley: I might be misunderstanding what some of the existing logic does, but I'm pretty sure that would be
`arg[0]: -Djavax.ssl.net.trustStore=C:\Program Files\security\truststore`
rather than splitting around the `=`.

And I agree, this is better. This still means that anyone passing a parameter that contains double quotes will need to change their existing bootstrap.properties, but I think that the odds of that are miniscule in comparison with people passing backslashes.

The only remaining question is how to escape a double-quote character. For the moment, I've done it by specifying the double quote twice i.e. `-Dmade.up.parameter="Something with spaces and ""double quotes"""` would become `-Dmade.up.parameter=Something with spaces and "double quotes"`. I can easily change it to escaping with a backslash (with double quote being the **only** character that can be escaped) if you think that would be better.

ideeley: OK, updated as discussed. Example `-Dmade.up.parameter="Something with spaces and \"double quotes\""` will be passed as `-Dmade.up.parameter=Something with spaces and "double quotes"`. Only double quotes can be escaped, and only inside a quote block.

ideeley: Forgot to confirm that `-Djavax.ssl.net.trustStore="C:\Program Files\security\truststore"` will still be passed as `-Djavax.ssl.net.trustStore=C:\Program Files\security\truststore`

modules/cli/pkg/launcher/jvmLauncher.go
```diff
@@ -786,11 +786,43 @@ func appendArgsBootstrapJvmLaunchOptions(args []string, bootstrapProperties prop
...
+						// If we've reached an unquoted space, that marks the end of the argument so
+						// we add what we've built so far to the list of args returned
+						args = append(args, argBuilder.String())
+						argBuilder.Reset()
+					} else {
+						argBuilder.WriteString(launchOptionChars[i])
+					}
+				} else {
+					if i < len(launchOptionChars) - 1 &&
+							launchOptionChars[i] == api.BOOTSTRAP_PROPERTY_NAME_LOCAL_JVM_LAUNCH_OPTIONS_ESCAPE &&
```
techcobweb, line 810: I think this logic is all tricky enough that a batch of unit tests would be good, to make sure that, given the properties as input, it harvests the correct arg string.

Can we have some unit tests please?

techcobweb: Especially when \ is the last character in the property... I suspect that it will go bang at that point with array limit exceeded ? Proof of that would be good.

techcobweb:
Note: https://docs.oracle.com/cd/E23095_01/Platform.93/ATGProgGuide/html/s0204propertiesfileformat01.html

Also says that the properties file can have multi-line content, so `\` characters ending the line should be treated as a continuation character. Can it handle that too ?

modules/cli/pkg/launcher/jvmLauncher.go
ideeley: No, the check for i < len(launchOptionChars) - 1 on line 809 catches that.

techcobweb: Prove it with unit tests.

Unit tests are partly there to make sure that if anyone changes this logic (tries to fix it), then the logic still behaves the same as it used to.

Also, the unit tests express a contract of what's valid, and what's not valid in terms of behavior.

ideeley: Yes, not trying to get out of that - just wanted to point out that that specific case was covered by the logic.

ideeley:
No, it can't handle multi-line content. It couldn't before either (and yes, I **have** tested this) so I haven't broken anything. The `\` and new line characters are processed as part of the property string, exactly as they were before. I'm not fixing that.

jadecarino (approved):
I have reviewed just the added unit tests given that Mike has reviewed the rest of the changes. Unit tests look good to me, thanks for adding these, happy to approve. Branch needs updating please then I can merge.

jadecarino:
> No, it can't handle multi-line content. It couldn't before either (and yes, I **have** tested this) so I haven't broken anything. The `\` and new line characters are processed as part of the property string, exactly as they were before. I'm not fixing that.

Yes just to continue this discussion, I've just tried this myself and for me I was able to set a CPS property to have multi-line with `\`, e.g.:
```
zos.cluster.DEFAULT.images=\
                           IMAGE1,\
                           IMAGE2
```
and that worked fine and became `zos.cluster.DEFAULT.images=IMAGE1,IMAGE2`.

But I just tried using a multi-line value for `galasactl.jvm.local.launch.options` and then the JVM failed to start. So looks like we have a bug there @techcobweb if that property should be able to support multi-line. Ian's changes haven't affected this.

ideeley:
I was nervous about looking at the multi-line bug on launch options as there are (now) design issues to consider. For example, what do we do if there's a line ending in the middle of an open quoted block?

modules/cli/pkg/launcher/jvmLauncher.go
jadecarino: Unit tests added in d547612dc4f48b2859a261eee9f2a6eaaa7c7650

modules/cli/pkg/embedded/templates/galasahome/bootstrap.properties
jadecarino: Resolving this thread as behaviour discussed has been implemented and unit tested, and changes were approved by @techcobweb

PR #350, opened 2025-06-19
Engine controller run monitoring threads wait for etcd and RAS pods to be ready before proceeding

eamansour:
## Why?
Builds off changes in https://github.com/galasa-dev/galasa/pull/347
Related to story https://github.com/galasa-dev/projectmanagement/issues/2235

## Changes
- All run-related monitoring threads that run within the engine controller pod now wait for the etcd and RAS pods to be up and ready before proceeding with their monitoring processing logic, rather than have them attempt to perform update operations and fail continuously
- Updated unit tests to account for the additional check at the start of each thread's `run()` method


jadecarino (approved):
Great stuff :)

PR #353, opened 2025-06-20
Iss2214 - Set `runLogStart` and `runLogEnd` in test structure for methods in a test class

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2214

Set runLogStart and runLogEnd in test structure for methods in a test class allowing users to easily find the log output relating to a specific method, either manually or in future, from the Galasa UI.

**Note:** the `runLogStart` and `runLogEnd` for `@Before` and `@After` methods are not being set and are still 0, apart from for the last `@Test` method. This is due to a general defect in which the `@Before` and `@After` methods apart from the last ones are not being updated at all in the test structure. A defect has been raised for this [here](https://github.com/galasa-dev/projectmanagement/issues/2291).

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
@@ -27,6 +27,8 @@ import dev.galasa.framework.spi.ConfigurationPropertyStoreException;
 import dev.galasa.framework.spi.DynamicStatusStoreException;
 import dev.galasa.framework.spi.FrameworkException;
 import dev.galasa.framework.spi.IDynamicStatusStoreService;
+import dev.galasa.framework.spi.IFramework;
+import dev.galasa.framework.spi.IResultArchiveStore;
 import dev.galasa.framework.spi.Result;
 import dev.galasa.framework.spi.teststructure.TestMethod;
 import dev.galasa.framework.spi.teststructure.TestStructure;
@@ -229,7 +231,7 @@ public class TestClassWrapper {
      */
     private void runGenericMethods(@NotNull ITestRunManagers managers, ArrayList<GenericMethodWrapper> genericMethods) throws TestRunException {
         for (GenericMethodWrapper genericMethod : genericMethods) {
-            genericMethod.invoke(managers, this.testClassObject, null);
+            genericMethod.invoke(managers, this.testClassObject, null, this);
             // Set the result so far after every generic method
             Result beforeClassMethodResult = genericMethod.getResult();
             setResult(beforeClassMethodResult, managers);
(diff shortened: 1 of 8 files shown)


modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
```diff
@@ -410,5 +412,24 @@ protected boolean isContinueOnTestFailureSet() {
         }
         return isContinueOnTestFailureSet;
     }
+    
+    protected IFramework getFramework() {
+        return this.testRunner.getFramework();
+    }
+
+    protected int getRunLogLines() {
```
eamansour: Minor - I initially confused `getRunLogLines()` to return the actual lines from the run log, so I wonder if this should be renamed to something like `getNumberOfRunLogLines()` or `getRunLogLineCount()` (feel free to go with anything you think would make more sense!)

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/ras/directory/DirectoryResultArchiveStoreService.java
```diff
@@ -50,10 +50,12 @@ public class DirectoryResultArchiveStoreService implements IResultArchiveStoreSe
     private Path                           testStructureFile;
     private Path                           runLog;
 
-    private final GalasaGson                     gson     = new GalasaGson();
+    private final GalasaGson               gson     = new GalasaGson();
 
     private DirectoryRASFileSystemProvider provider;
 
+    private String                         message;
```
eamansour: Minor - instead of `message`, this could be renamed to `runLogContent` to make it more descriptive? (again feel free to go with whatever you think would make the most sense)

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/IResultArchiveStore.java
```diff
@@ -55,6 +55,14 @@ public interface IResultArchiveStore {
      */
     void writeLog(@NotNull List<String> messages) throws ResultArchiveStoreException;
 
+    /**
+     * Retrieves the log so far from the RAS as a String so we can
+     * work out the start and end line numbers of the run log for 
+     * each method in a test class.
+     * @return the run log as a String at any given time in a test.
+     */
+    default String retrieveLog() { return ""; };
```
eamansour, line 64: Should we remove the `default` implementation for this? That way, we'll know that all IResultArchiveStore implementations have overridden this function.

At the moment, the CouchDB RAS extension doesn't seem to override this, so remote runs won't set the runLogStart and runLogEnd to anything other than 0.

eamansour (approved):
A few minor comments. Happy for you to implement the CouchDB extension's`retrieveLog` method in a separate PR if you'd prefer, so approved.

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
jadecarino: Great point, thank you, fixed in https://github.com/galasa-dev/galasa/pull/353/commits/03c6c7f4f540ef624bd8935d1ddf9cbd6ecbc9c9

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/ras/directory/DirectoryResultArchiveStoreService.java
jadecarino: Yes that makes sense thank you, fixed in https://github.com/galasa-dev/galasa/pull/353/commits/03c6c7f4f540ef624bd8935d1ddf9cbd6ecbc9c9

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/IResultArchiveStore.java
jadecarino: Yes sure! I will do this in a separate PR so that I can get this first part delivered for local runs now.

PR #355, opened 2025-06-25
Requestor id now case insensitive

aashir21:
# Why?

- Refer to https://github.com/galasa-dev/projectmanagement/issues/2252


modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/ras/RasSearchCriteriaRequestor.java
```diff
@@ -27,7 +27,7 @@ public boolean criteriaMatched(@NotNull TestStructure structure) {
 		
 		if(requestors != null) {
 			for(String requestor : requestors) {
-				if(requestor.equals(structure.getRequestor())){
+				if(requestor.equalsIgnoreCase(structure.getRequestor())){
```
eamansour: `criteriaMatched` isn't used by the CouchDB extension, only the local directory-based RAS. This change will have no impact on querying runs in CouchDB unfortunately.

As mentioned in https://github.com/galasa-dev/projectmanagement/issues/2252#issuecomment-2984710591, we need to modify the `/ras/runs` endpoint to match a given requestor with a username as stored in the auth store. That matching process should be case-insensitive, and once a match is found, we use the username from the auth store for querying runs.

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -215,7 +215,7 @@ private List<RasRunResult> getRunsByIds(List<String> runIds, boolean isMethodDet
         return runs;
     }
 
-	private List<IRasSearchCriteria> getCriteria(RasQueryParameters queryParams) throws InternalServletException {
+	private List<IRasSearchCriteria> getCriteria(RasQueryParameters queryParams) throws InternalServletException, ResultArchiveStoreException {
```
eamansour: Is this new ResultArchiveStoreException caught somewhere? Might be worth having a try-catch to only have an InternalServletException being thrown

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -246,8 +246,11 @@ private List<IRasSearchCriteria> getCriteria(RasQueryParameters queryParams) thr
 			critList.add(toCriteria);
 		}
 		if (requestor != null && !requestor.isEmpty()) {
-			RasSearchCriteriaRequestor requestorCriteria = new RasSearchCriteriaRequestor(requestor);
-			critList.add(requestorCriteria);
+			String matchedRequestor = findMatchingRequestor(requestor);
+			if(matchedRequestor != null && matchedRequestor.isBlank()){
```
eamansour: This if-statement will only be entered if a blank requestor is matched (`&& matchedRequestor.isBlank()`)

Would've thought this could be caught in a unit test, so maybe we're missing one that goes through this code?

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -312,6 +315,23 @@ private String buildResponseBody(List<RasRunResult> runs, int pageNum, int pageS
         return gson.toJson(runsPage);
 	}
 
+	private String findMatchingRequestor(String requestor) throws ResultArchiveStoreException, InternalServletException {
+		String matchedRequestor = null;
+
+		// Get all requestors in the service
+		List<String> requestorsList = getRequestors();
```
eamansour: Assuming this `getRequestors()` could throw a ResultArchiveStoreException, maybe we could have a try-catch and throw an appropriate InternalServletException instead?

eamansour (commented):
Looking better! Spotted a few more minor improvements that can be made

eamansour (approved):

PR #359, opened 2025-07-02
Iss2214 - Update and retrieve run log line count in Directory and CouchDB RAS store

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2214

## Changes

- Update a class variable with the run log line count in the Directory and CouchDB RAS stores.
- This can then be retrieved from the RAS through the framework, at the start and end of each method invocation, to place the start run log line and end run log line in the test structure.
- This PR includes updates to the changes delivered in #353 to the Framework, as well as new changes to the Extensions module.


modules/extensions/galasa-extensions-parent/dev.galasa.ras.couchdb/src/main/java/dev/galasa/ras/couchdb/internal/CouchdbRasStore.java
```diff
@@ -90,9 +90,11 @@ public class CouchdbRasStore extends CouchdbStore implements IResultArchiveStore
     private String                             artifactDocumentRev;
 
     private TestStructure                      lastTestStructure;
-    private ITimeService timeService ;
+    private ITimeService                       timeService ;
 
-    private LogFactory logFactory;
+    private LogFactory                         logFactory;
+
+    private int                                runLogLineCount;
```
eamansour: Might be a non-issue, but could the run log exceed the max value for integers?

eamansour (approved):

modules/extensions/galasa-extensions-parent/dev.galasa.ras.couchdb/src/main/java/dev/galasa/ras/couchdb/internal/CouchdbRasStore.java
jadecarino: Not sure I don't think we've personally had a run log that large, but other users potentially could. I will change this to a `long` in case.

jadecarino: Updated in commit https://github.com/galasa-dev/galasa/pull/359/commits/167fce9244c8187c1f86bcba7b88dde7f9b3750f

eamansour (approved):
Looks good!

PR #361, opened 2025-07-02
mcobbett null pointer ex in 3270 cleanup 2271

techcobweb:
# Why ?
See [Started seeing irregular NullPointerException's in zos3270 manager disconnect() #2271](https://github.com/galasa-dev/projectmanagement/issues/2271)

I think there is a race condition which gets hit occasionally, causing a NPE.

Trying to add a lock to prevent this race condition from occurring.

- Added a section in the CONTRIBUTING.md to tell contributors to sign their commits
- Terminal should protect against race condition with networkThread variable being nulled-out.

### DIFF
--- a/modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/spi/Terminal.java
+++ b/modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/spi/Terminal.java
@@ -33,18 +33,22 @@ import dev.galasa.zos3270.internal.comms.NetworkThread;
 
 public class Terminal implements ITerminal {
 
+    // The max amount of time this thread will wait around for the network thread to finish after
+    // we have closed the network. Units are milliseconds. 
+    public static final long MAX_MILLISECS_TO_WAIT_FOR_NETWORK_THREAD_TO_FINISH = 20 * 1000 ; // 20 seconds
+
 	public ITextScannerManagerSpi textScan;
     private final Screen  screen;
     private final Network network;
     private final String  id;
-    private NetworkThread networkThread;
-    private boolean connected = false;
+    private volatile NetworkThread networkThread;
+    private volatile boolean connected = false;
 
     private int           defaultWaitTime = 120_000;
 
     private Log           logger          = LogFactory.getLog(getClass());
     
-    private boolean       autoReconnect   = false;
+    private volatile boolean       autoReconnect   = false;
     
     private List<String>  deviceTypes;
     private String        requestedDeviceName;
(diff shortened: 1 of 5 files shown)


modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/spi/Terminal.java
```diff
@@ -155,16 +181,36 @@ public void disconnect() throws TerminalInterruptedException {
...
+        }
+
+        logger.trace("disconnect: Thread "+Thread.currentThread().hashCode()+" is about to request the lock for the network thread");
+        synchronized(networkThreadLock) {
+            logger.trace("disconnect: Thread "+Thread.currentThread().hashCode()+" has acquired the lock for the network thread");
+            try {
+                if (networkThread != null) {
+                    try {
+                        logger.trace("Waiting for network thread to end...");
+                        networkThread.join(MAX_WAIT_MILLISECONDS_FOR_NETWORK_THREAD_TO_DIE);
```
eamansour: Starting to wonder whether the networkThread does something before it exits that also tries to acquire the same lock as this method. If that happens, then both will enter a waiting state.

Once this 'timeout' is reached and the lock is released, the networkThread will continue, could that lead to another race condition?


Probably need to go through the disconnect-related code paths to see what actually happens.

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/spi/Terminal.java
```diff
@@ -155,16 +181,36 @@ public void disconnect() throws TerminalInterruptedException {
...
+            logger.trace("connect: Thread "+Thread.currentThread().hashCode()+" has acquired the lock for the network thread");
             try {
-                logger.trace("Waiting for network thread to end...");
-                networkThread.join();
-                logger.trace("Network thread ended OK");
-            } catch (InterruptedException e) {
-                logger.warn("Interrupted while waiting for network thread to end. Network thread might still be active.");
-                throw new TerminalInterruptedException("Join of the network thread was interrupted",e);
+                networkThread = new NetworkThread(this, screen, network, network.getInputStream(), this.deviceTypes);
+                networkThread.start();
```
eamansour: Curious about this (unsure if it's intentional) - do we need to start a new networkThread when disconnecting?

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/spi/Terminal.java
eamansour: Also not sure how we can unit test this reliably - can we spin up a few network threads in a test and ask a terminal to disconnect?

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/spi/Terminal.java
techcobweb: It looked like there was re-connect logic in there also. So I thought a belt-and-braces approach to protect all uses of the variable consistently would be safer.

techcobweb:
Backing out those changes and looking at it again.

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/spi/Terminal.java
techcobweb: Backed-out the locking.

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/spi/Terminal.java
techcobweb: Backed-out the locking.

eamansour (approved):
Looks okay - hard to tell when it comes to this bit of the code! Approved given that the 3270 IVT passes.

PR #362, opened 2025-07-03
Iss2257 - Requestor shown by galasactl in `runs get`, generated reports and CLI output should be authenticated user

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2257

## Changes

- Updated the summary output that is shown during/at the end of a runs submit, to show the `requestor` as the user who is authenticated to the Galasa API Server and therefore is the `requestor` placed in the run's test structure, instead of the current system user who ran `galasactl`.
- The change above also shows the correct `requestor` now too in the generated JSON, YAML and JUnit reports.
- Refactored the submitter/remote launcher code to check what the current system user is and if it matches the authenticated user that is the actual `requestor`, and logs a different message to the console based on this.


modules/cli/pkg/launcher/remoteLauncher.go
```diff
@@ -85,14 +85,12 @@ func (launcher *RemoteLauncher) SubmitTestRun(
 	testRunRequest := galasaapi.NewTestRunRequest()
 	testRunRequest.SetClassNames(classNames)
 	testRunRequest.SetRequestorType(requestType)
-	testRunRequest.SetRequestor(requestor)
+	testRunRequest.SetRequestor(systemUser)
```
eamansour: If the API server uses the requestor in the JWT, do we still need to set the requestor in the request payload?

eamansour (approved):

modules/cli/pkg/launcher/remoteLauncher.go
jadecarino: Yes I thought this, it does seem a little useless as its essentially ignored in the test run request. I can remove it for clarity so it doesn't look like it's being used.

jadecarino: Fixed in 1d1e00f96513ca8cdd74b42e40eae0751e30f40b

PR #365, opened 2025-07-16
Iss2312 - Allow max gRPC message size to be set from Helm Chart value

jadecarino:
## Why?

Allow max gRPC message size to be set from Helm Chart value to avoid the error `RESOURCE_EXHAUSTED: gRPC message exceeds maximum size`.

## Changes

- Retrieve an environment variable for the MAX_GRPC_MESSAGE_SIZE, default if not provided, and check valid value for an int. 
- Pass through to the Etcd3DynamicStatusStore constructor to be used when setting up the KVClient.
- Unit tests added for Etcd3DynamicStatusStore constructor.


jadecarino:
We are no longer making this change so closing this PR.

PR #368, opened 2025-07-17
Iss2312 - Get only keys from ETCD for certain calls (efficiency)

jadecarino:
## Why?

## Changes

- refactor: Get only the keys from ETCD rather than key/values for run polling/submitting to avoid large data being sent back.
   - There is no need to get all the key/values back from the DSS for all keys with a prefix `dss.framework.run` as all thats used is the keys, to then create a bunch of IRun structures. This change will hopefully avoid the request reported in the issue where the max message size is being reached.


modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStore.java
```diff
@@ -305,31 +305,66 @@ public String get(@NotNull String key) throws DynamicStatusStoreException {
     @Override
     public @NotNull Map<String, String> getPrefix(@NotNull String keyPrefix) throws DynamicStatusStoreException {
 
-        logger.debug("Etcd extension getting property with a prefix of "+keyPrefix);
+        Map<String, String> keyValues = new HashMap<>();
+
+        List<KeyValue> kvs = getPropertiesFromETCDWithPrefix(keyPrefix, false);
```
techcobweb: a constant symbolic would be clearer to read

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStore.java
```diff
@@ -305,31 +305,66 @@ public String get(@NotNull String key) throws DynamicStatusStoreException {
...
+
+        if (kvs.isEmpty()) {
+            return new HashMap<>();
```
techcobweb: just drop through to the return statement ?

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStore.java
```diff
@@ -305,31 +305,66 @@ public String get(@NotNull String key) throws DynamicStatusStoreException {
...
+
+    /**
+     * A get of all keys that start with a specified prefix. They are returned in a list.
+     * 
+     * @param keyPrefix - the prefix for any key(s)
+     * @return A list of keys
+     * @throws DynamicStatusStoreException A failure occurred.
+     */
+    @Override
+    public List<String> getPrefixKeysOnly(@NotNull String keyPrefix) throws DynamicStatusStoreException {
```
techcobweb: Should this be returning a collection<String> rather than a list ? depends if there is some sorted order I guess.

techcobweb (approved):

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStore.java
jadecarino: Fixed in https://github.com/galasa-dev/galasa/pull/368/commits/46b1cea9dbf8df14350c52a8e79d644341bd16b5

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStore.java
jadecarino: Fixed in https://github.com/galasa-dev/galasa/pull/368/commits/2e55736c18e37ee77b795a58c9766cbba35954ea

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStore.java
jadecarino: Fixed in https://github.com/galasa-dev/galasa/pull/368/commits/2e55736c18e37ee77b795a58c9766cbba35954ea

PR #372, opened 2025-07-22
build: Add retry loop when sending requests to argocd.galasa.dev in case of intermittent 502 error

jadecarino:
Signed-off-by: Jade Carino <carino_jade@yahoo.co.uk>


.github/workflows/obr.yaml
```diff
@@ -279,15 +279,41 @@ jobs:
...
+          for i in {1..10}; do
+            docker run \
+            --env ARGOCD_AUTH_TOKEN=${{ env.ARGOCD_AUTH_TOKEN }} \
+            --rm ghcr.io/${{ env.NAMESPACE }}/argocdcli:main \
+            app actions run ${{ env.BRANCH }}-maven-repos restart \
+            --kind Deployment \
+            --resource-name obr-${{ env.BRANCH }} \
+            --server argocd.galasa.dev \
+            --grpc-web \
+            && exit 0 || sleep 10
```
techcobweb, line 291: shouldn't the sleep 10 be on a separate line to the && exit 0 ? rather than OR-ed in ?

techcobweb (approved):

techcobweb:
Only one comment, but approved anyway as it will likely work as you wish I think. Just not 100% sure.

.github/workflows/obr.yaml
jadecarino: I have tested this on my fork and it works as expected so I think we should be okay

PR #374, opened 2025-07-23
Minor changes to make local builds work better

techcobweb:

# Why ?
- detect-secrets didn't fail if the version was less than 3.11.0 We should force people to upgrade them.
- readme didn't contain a list of pre-reqs which we need
- docs shouldn't need sourceMaven environment variable set to build locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,20 @@ This is the main source code repository for the Galasa open source project.
 - [`modules`](./modules/) - The code
 - [`tools`](./tools/) - Build tools and useful scripts
 
-## building locally
 
+
+## Building locally
+
+### Pre-reqs
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -5,8 +5,20 @@ This is the main source code repository for the Galasa open source project.
...
 - [`tools`](./tools/) - Build tools and useful scripts
 
-## building locally
 
+
+## Building locally
+
+### Pre-reqs
+
+These are the versions of software which are known to work:
```
James-Cocker, line 14: dont like this

James-Cocker: thanks for fixing.

James-Cocker (approved):

PR #377, opened 2025-07-24
Making sure the the @AfterClass methods are executed even if a test method throws an exception

techcobweb:

# Why ?

- Noticed by code inspection that a test which throws an exception when running the test supresses any calls to the @AfterClass
- Also, that it supresses any calls to tell the managers that the test has finished.
- Re-factored the testClassWrapper to make it easier to unit test


jadecarino (approved):
Changes look good, approved!

I wasn't sure if you wanted me to merge this on your behalf, or if you still had changes you wanted to add to this change set, so I'm leaving this unmerged.

PR #378, opened 2025-07-28
Fix bug where structure.json was not being updated for all `Before`/`After` methods

jadecarino:
Signed-off-by: Jade Carino <carino_jade@yahoo.co.uk>

## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2291

## Changes

- Fixed bug where the test structure was not updated for all `Before` and `After` methods, only the ones belonging to the last `Test` method to execute. This is because the same ArrayList of `GenericMethodWrapper`s was being pointed to for each `TestMethodWrapper` and so the structure would be re-initialised (set back to empty) each test method, with only the last Before and After having updates saved to the test structure.
- Renamed some variables for readability. As `TestMethodWrapper` and `GenericMethodWrapper` contain similar code, it was  sometimes confusing to read.
- Split the method originally called `getStructure` into two separate methods: an `initialise` which creates a new test structure and sets that as a class variable, and a `get` which returns that class variable. It didn't make sense to me that the original `getStructure` was doing two things, both setting and getting something.
- More refactoring for readability.

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
@@ -125,8 +125,19 @@ public class TestClassWrapper {
 
         // *** Build the wrappers for the test methods
         for (Method method : temporaryTestMethods) {
-            this.testMethods
-            .add(new TestMethodWrapper(method, this.testClass, temporaryBeforeMethods, temporaryAfterMethods));
+
+            ArrayList<GenericMethodWrapper> beforesForThisTestMethod = new ArrayList<>();
+            for (GenericMethodWrapper before : temporaryBeforeMethods) {
+                beforesForThisTestMethod.add(before.createCopyGenericMethodWrapper());
+            }
+
+            ArrayList<GenericMethodWrapper> afterMethodsForThisTestMethod = new ArrayList<>();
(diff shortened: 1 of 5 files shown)


modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
```diff
@@ -144,24 +144,41 @@ public void parseTestClass() throws TestRunException {
...
+                beforesForThisTestMethod.add(before.createCopyGenericMethodWrapper());
+            }
+
+            ArrayList<GenericMethodWrapper> afterMethodsForThisTestMethod = new ArrayList<>();
+            for (GenericMethodWrapper after : temporaryAfterMethods) {
+                afterMethodsForThisTestMethod.add(after.createCopyGenericMethodWrapper());
+            }
+
+            TestMethodWrapper testMethodWrapper = new TestMethodWrapper(method, this.testClass, beforesForThisTestMethod, afterMethodsForThisTestMethod);
+            this.testMethods.add(testMethodWrapper);
```
jadecarino, line 159: This is the fix for why only the Before and After methods of the last Test method had updated fields in the structure.json.

Each TestMethodWrapper was pointing at the same array of Befores/Afters and so the first Befores/Afters were being set back to their original state in `TestMethodWrapper.initialiseTestMethodStructure()` and only the last Befores/Afters had updated fields in the structure.json.

techcobweb (approved):

PR #379, opened 2025-07-30
mcobbett test pods realise they are cancelled 2

techcobweb:
# Why ?
See [Cancelling of run groups not working as expected #2317](https://github.com/galasa-dev/projectmanagement/issues/2317)

- [x] Test run checks before it starts invoking managers whether it's already been interrupted.
- [x] Test run checks between test methods whether it's been interrupted, and stops calling more test methods
- [x] Refactored so TestClassWrapper takes a RAS reference rather than a framework reference, and then gets the RAS from the Framework. Principle if least information.

This is test runner code so more than the normal level of review and re-testing will be needed before merging.

Tested it on minikube and I can see it interrupt a test between methods.

- [x] Remove 'Please' from some error messages.
- [x] Grouped together all the dss run suffixes into an enum to save mis-spellings.


modules/managers/galasa-managers-parent/galasa-managers-cicsts-parent/dev.galasa.cicsts.manager/src/main/java/dev/galasa/cicsts/internal/CicstsDefaultLogonProvider.java
```diff
@@ -134,7 +134,7 @@ private void waitForLoggedOnText(ICicsTerminal cicsTerminal) throws CicstsManage
 
         String[] pass = { "Sign-on is complete" };
         String[] fail = { 
-            "Your password has expired. Please type your new password.",
+            "Your password has expired. Type your new password.",
```
eamansour: This message might come from the server since the terminal waits for it to appear. I think this should be changed back to avoid any issues for tests using this manager.

modules/managers/galasa-managers-parent/galasa-managers-testingtools-parent/dev.galasa.sdv.manager/src/main/java/dev/galasa/sdv/internal/SdvUserImpl.java
```diff
@@ -136,7 +136,7 @@ public void setNotRecording() {
     private void waitForLoggedOnText(ICicsTerminal cicsTerminal) throws CicstsManagerException {
 
         String[] pass = {"Sign-on is complete"};
-        String[] fail = {"Your password has expired. Please type your new password.",
+        String[] fail = {"Your password has expired. Type your new password.",
```
eamansour: Same here - this could come from the server, so I wonder if changing this could cause issues?

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestMethodWrapper.java
```diff
@@ -79,6 +79,12 @@ public void invoke(
 
                 runBeforeMethods(managers, testClassObject, testClassWrapper);
 
+                // TODO After this point, should we invoke after methods even if the test method fails with an exception ?
+                // Seems that an exception will skip lots of things otherwise.
+                // Failures like an exception being thrown from the testcase method itself don't get thrown up to this point, only if 
+                // there is a bad framework issue such as the dss being unavailable or something. In which case all bets are probably off anyway.
+                // Managers will be invited to clean up even if we throw an exception out of this method.
```
eamansour: I would assume that the after methods would be invoked to allow some cleanup to take place if possible (e.g. like resetting a terminal screen), but yes, with wider framework issues, exiting out and letting the managers clean up sounds fine.

I think this is OK.

eamansour (approved):
A couple of minor comments but overall looks good - approved

modules/managers/galasa-managers-parent/galasa-managers-cicsts-parent/dev.galasa.cicsts.manager/src/main/java/dev/galasa/cicsts/internal/CicstsDefaultLogonProvider.java
techcobweb: Reverted

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestMethodWrapper.java
techcobweb: OK. I left the logic as-is. Removed the TODO and made it a comment.
Bad failures like framework exceptions coming from the managers will fail here.

modules/managers/galasa-managers-parent/galasa-managers-testingtools-parent/dev.galasa.sdv.manager/src/main/java/dev/galasa/sdv/internal/SdvUserImpl.java
techcobweb: Reverted

PR #382, opened 2025-08-07
New test run gets put into the queued state in the RAS record when it is first created.

techcobweb:

# Why ?
See [Cancelling of run groups not working as expected #2317](https://github.com/galasa-dev/projectmanagement/issues/2317)

When the test is launched, the record which gets placed into RAS initially should have the Queued state, with the queued time.

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.api.runs/src/main/java/dev/galasa/framework/api/runs/common/GroupRuns.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework.api.runs/src/main/java/dev/galasa/framework/api/runs/common/GroupRuns.java
@@ -30,6 +30,7 @@ import dev.galasa.framework.spi.ResultArchiveStoreException;
 import dev.galasa.framework.spi.rbac.RBACException;
 import dev.galasa.framework.spi.teststructure.TestStructure;
 import dev.galasa.framework.spi.utils.GalasaGson;
+import dev.galasa.framework.spi.utils.ITimeService;
 
 import static dev.galasa.framework.api.common.ServletErrorMessage.*;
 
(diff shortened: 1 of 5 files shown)


modules/framework/galasa-parent/dev.galasa.framework.api.runs/src/main/java/dev/galasa/framework/api/runs/common/GroupRuns.java
```diff
@@ -146,8 +148,10 @@ public ScheduleStatus scheduleRun(ScheduleRequest request, String groupName, Str
         return status;
     }
 
-    private void createRunRasRecord(IRun newRun) throws InternalServletException {
+    protected void createRunRasRecord(IRun newRun, IResultArchiveStore rasStore, ITimeService timeService) throws InternalServletException {
         TestStructure newTestStructure = newRun.toTestStructure();
+        newTestStructure.setStatus("queued");
```
eamansour: Instead of a literal for "queued", could we use the existing `TestRunLifecycleStatus.QUEUED` value?

```suggestion
        newTestStructure.setStatus(TestRunLifecycleStatus.QUEUED.toString());
```

eamansour (approved):
One minor suggestion but approved

modules/framework/galasa-parent/dev.galasa.framework.api.runs/src/main/java/dev/galasa/framework/api/runs/common/GroupRuns.java
techcobweb: Yea. Wondering what it will take to use this throughout the code now.
If we use it everywhere then there's no chance of bad strings...etc.
I think it should be strongly typed. Worry about it breaking stuff though.
All the time it's a string literal, people will be supplying string literals.

techcobweb: Fixed.

PR #396, opened 2025-08-20
openapi REST interface - documented that the requestor field supplied when a test is launched is not used

techcobweb:

# Why ?
See issue: [requestor is no longer used, but its still in the docs and can be provided #2385](https://github.com/galasa-dev/projectmanagement/issues/2385)

Marking the requestor field as deprecated in comments in the REST interface on the test launch API call.

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
+++ b/modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
@@ -2927,6 +2927,9 @@ components:
           type: string
         requestor:
           type: string
+          description: |-
+            The name of the user who submitted this test, in that it represents the owner 
+            of the Galasa personal access token used by the test submission REST call.
         stream:
           type: string
         repo:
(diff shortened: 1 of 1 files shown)


modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -3155,6 +3174,7 @@ components:
           description: A uuid created to identify this test run. Useful until this run has been allocated a test name, which is allocated once the test is launched.
         requestor:
           type: string
+          description: The Gaalasa user who requested this test to be executed.
```
Mark-J-Lawrence: typo

jadecarino (approved):
Approved assuming the typo Mark pointed out will be addressed

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
techcobweb: Fixed

PR #397, opened 2025-08-20
docs don't show how to create a stream using older properties set mechanism

techcobweb:

# Why ?
As a result of story [Docs: Streams need to be created using yaml #2334](https://github.com/galasa-dev/projectmanagement/issues/2334)

The old text described how to create a stream using properties. Now we refer to the existing docs which talks about doing it with yaml resource files.

I wasn't brave enough to try to combine all the information together under one Streams title.

But the docs are (slightly) better as a result I believe.

### DIFF
--- a/docs/content/docs/manage-ecosystem/test-streams.md
+++ b/docs/content/docs/manage-ecosystem/test-streams.md
@@ -20,18 +20,14 @@ The following diagram shows the relationship between the test code, test catalog
 ![test stream architecture:](test-streams-architecture.svg)
 
 
-## Creating and retrieving test stream components
+## Creating and retrieving a test stream
 
-The components of the `test.stream` property are set by using `galasactl properties set` command, as shown in the following example:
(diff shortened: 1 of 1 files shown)


docs/content/docs/manage-ecosystem/test-streams.md
```diff
@@ -20,18 +20,14 @@ The following diagram shows the relationship between the test code, test catalog
 ![test stream architecture:](test-streams-architecture.svg)
 
 
-## Creating and retrieving test stream components
+## Creating and retrieving a test stream
 
-The components of the `test.stream` property are set by using `galasactl properties set` command, as shown in the following example:
+The components of the `test.stream` property are set by using `galasactl resources apply -f {yaml-filename}` command.
+The `apply` sub-command will update the stream if it already exists, but you could use the `create` or `update` sub-command as an alternative
```
eamansour: Minor - missing a `.` to end this sentence.

eamansour (approved):
One minor comment but approved

PR #402, opened 2025-08-22
Encode spaces in RAS paths to stop an illegal character exception from being thrown

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2363

## Changes
- When a ResultArchiveStorePath is created, spaces in the path are now percent-encoded into `%20`
- Added unit tests to verify this functionality
- Added a test to the ZosFileManagerIVT to test that a unix file with spaces in its file path can be created and saved to the RAS without issue
- Updated existing RAS path tests to use a mock file system instead of the default (real) filesystem


jadecarino (approved):
Looks good to me!

PR #403, opened 2025-08-26
Add toJsonString() method to ITerminal interface and add a test to Zos3270IVT testing the new method

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2353

## Changes
- Added a new `toJsonString()` method to the 3270 manager's `ITerminal` interface which converts the current ITerminal instance into a string in JSON format
- Added a test to the Zos3270IVT test suite to exercise this new method

### DIFF
--- a/modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.common/src/main/java/dev/galasa/zos3270/common/screens/json/TerminalJsonTransform.java
+++ b/modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.common/src/main/java/dev/galasa/zos3270/common/screens/json/TerminalJsonTransform.java
@@ -6,8 +6,14 @@
 package dev.galasa.zos3270.common.screens.json;
 
 
+import java.util.ArrayList;
+import java.util.Map.Entry;
+
 import com.google.gson.Gson;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
 
 import dev.galasa.framework.spi.utils.GalasaGsonBuilder;
 import dev.galasa.zos3270.common.screens.Terminal;
@@ -19,22 +25,48 @@ public class TerminalJsonTransform {
 
     private Gson gson ;
 
+    public TerminalJsonTransform() {
+        this(true);
+    }
+
     public TerminalJsonTransform( boolean isPrettyPrinting ) {
         this.gson = new GalasaGsonBuilder(isPrettyPrinting).getGson();
     }
     
     public JsonObject toJsonObject(Terminal terminal) {
-        JsonObject json = (JsonObject) gson.toJsonTree(terminal);
-        return json;
+        return (JsonObject) gson.toJsonTree(terminal);
     }
     
     public String toJsonString(Terminal terminal) {
         JsonObject jsonObj = toJsonObject(terminal);
-        String jsonString = gson.toJson(jsonObj);
-        return jsonString;
+        stripFalseBooleans(jsonObj);
+        return gson.toJson(jsonObj);
     }
 
     public Terminal toTerminal(String tempJson) {
         return gson.fromJson(tempJson, Terminal.class);
+    }
+
+    private void stripFalseBooleans(JsonObject json) {
+        ArrayList<Entry<String, JsonElement>> entries = new ArrayList<>();
+        entries.addAll(json.entrySet());
+
+        for (Entry<String, JsonElement> entry : entries) {
+            JsonElement element = entry.getValue();
+
+            if (element.isJsonPrimitive() && ((JsonPrimitive) element).isBoolean()
+                    && !((JsonPrimitive) element).getAsBoolean()) {
+                json.remove(entry.getKey());
+            } else if (element.isJsonObject()) {
+                stripFalseBooleans((JsonObject) element);
+            } else if (element.isJsonArray()) {
+                JsonArray array = (JsonArray) element;
+                for (int i = 0; i < array.size(); i++) {
+                    if (array.get(i).isJsonObject()) {
+                        stripFalseBooleans((JsonObject) array.get(i));
+                    }
+                }
+            }
+        }
     }
 }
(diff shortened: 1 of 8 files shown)


modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.common/src/main/java/dev/galasa/zos3270/common/screens/json/TerminalJsonTransform.java
```diff
@@ -19,22 +25,48 @@ public class TerminalJsonTransform {
...
-        return jsonString;
+        stripFalseBooleans(jsonObj);
+        return gson.toJson(jsonObj);
     }
 
     public Terminal toTerminal(String tempJson) {
         return gson.fromJson(tempJson, Terminal.class);
     }
+
+    private void stripFalseBooleans(JsonObject json) {
```
eamansour, line 50: FYI - this function has been moved from https://github.com/galasa-dev/galasa/pull/403/files#diff-f5c5977d7099b7b3e11a6d686c6e86adbfb503c3fc3b959e41b88583790f5c41L417-L438

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/spi/Terminal.java
```diff
@@ -779,5 +783,17 @@ public Highlight retrieveHighlightAtPosition(int row, int col) throws Zos3270Exc
...
+        String terminalJsonStr = null;
+        if (currentTerminal != null) {
+            terminalJsonStr = this.terminalJsonTransform.toJsonString(currentTerminal);
+        }
+        return terminalJsonStr;
+    }
+
+    protected void setCurrentTerminal(dev.galasa.zos3270.common.screens.Terminal currentTerminal) {
+        this.currentTerminal = currentTerminal;
+    }
```
eamansour, line 797: The terminal screen info is populated in `Zos3270TerminalImpl`, which is a subclass of this `Terminal` class. To avoid leaving the `toJsonString()` method unimplemented in the `Terminal` class, I've added a `currentTerminal` field that can be updated by the subclass and then this `toJsonString()` method can simply convert the current terminal into a JSON string.

Happy to change this if you can think of a better way to avoid leaving `toJsonString()` unimplemented in this class!

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.common/src/main/java/dev/galasa/zos3270/common/screens/json/TerminalJsonTransform.java
jadecarino: Yep that makes good sense :)

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/test/java/dev/galasa/zos3270/spi/TerminalTest.java
```diff
@@ -45,5 +50,106 @@ public boolean connectClient() throws NetworkException {
         textScan, codePage, network);
 
     }
+
+    @Test
+    public void testConvertNullTerminalToJsonReturnsNull() throws Exception {
```
jadecarino: The name of this method confused me a little, as it doesn't appear that the Terminal itself is null, rather the screen should be at this point? I'm not certain though. From L75 it looks to me like Terminal is not null.

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/test/java/dev/galasa/zos3270/spi/TerminalTest.java
```diff
@@ -45,5 +50,106 @@ public boolean connectClient() throws NetworkException {
...
+        dev.galasa.zos3270.common.screens.Terminal rasTerminal = new dev.galasa.zos3270.common.screens.Terminal(terminalId, runId, sequenceNo, primarySize);
+        TerminalImage image = new TerminalImage(0, terminalId, isInbound, type, aid, primarySize, cursorColumn, cursorRow);
+
+        TerminalField field = new TerminalField(0, 0, false, false, false,
+            true, false, false, false, Colour.BLUE.getLetter(),
+                Colour.DEFAULT.getLetter(), Highlight.DEFAULT.getLetter()
+        );
+        FieldContents fieldContents = new FieldContents(fieldChars);
+        field.getContents().add(fieldContents);
+        image.getFields().add(field);
```
jadecarino: This might benefit from some comments for those who aren't as familiar with 3270, as I'm not sure what's going on here and why the TerminalField for example is being set up like that, is there a reason or is it arbitrary? Anyway, a few short comments might help here.

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/main/java/dev/galasa/zos3270/spi/Terminal.java
jadecarino: No this seems reasonable to me. Thanks for the extra info!

jadecarino (approved):
Approved but some clarification on the method name and some comments if you think that would be useful would be good if you want to add this. Thanks!

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/test/java/dev/galasa/zos3270/spi/TerminalTest.java
eamansour: Agreed, it's a bit confusing because we have a Terminal class which represents the actual terminal and its functionality, but we also have another Terminal class that's just a pojo representing a terminal screen 😅 

Addressed in https://github.com/galasa-dev/galasa/pull/403/commits/17e34c1ca77f045b63db3909d5f5eeba45654f72

modules/managers/galasa-managers-parent/galasa-managers-zos-parent/dev.galasa.zos3270.manager/src/test/java/dev/galasa/zos3270/spi/TerminalTest.java
eamansour: Thanks - this setup is all arbitrary, it just populates the fields in a terminal screen for the unit test. Added comments in https://github.com/galasa-dev/galasa/pull/403/commits/17e34c1ca77f045b63db3909d5f5eeba45654f72

PR #405, opened 2025-08-27
Iss2369 - Uplift Gradle version to 9.0.0 and update build processes

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2369

## Changes

- Uplift Gradle version used in GitHub Actions workflows to 9.0.0 from 8.9.
- Uplift biz.aQute.bnd dependencies to version 7.1.0 as this version of the plugin is compatible with Gradle 9.0.0.
- Remove use of -b flag to point to location of build.gradle file as this has been removed in Gradle 9.
- Replace sourceCompatibility and targetCompatibility with Java toolchains as this is the recommended way to set the Java version and setting sourceCompatibility and targetCompatibility is not supported in Gradle 9.
- Gradle 9 now throws an error if it detects test source files in src/test/java but these files do not have any runnable test cases. This PR introduces an excludes section to the build.gradle files of projects that have test files with no runnable tests to exclude Gradle from parsing them.
- junit-platform-launcher has been added to the SDV Manager as it requires a JUnit runtime (Gradle 9 throws an error now if this is missing).
- Missing slash added in Framework's build-locally.sh as Gradle 9 now throws an error.

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.api.launcher/build.gradle
+++ b/modules/framework/galasa-parent/dev.galasa.framework.api.launcher/build.gradle
@@ -9,6 +9,12 @@ dependencies {
     implementation project(':dev.galasa.framework')
 }
 
+test {
+    // Exclude the LauncherTest.java class
+    // As it has no runnable Tests currently, this causes an error in Gradle 9.0.0.
+    exclude '**/LauncherTest.class'
+}
+
 // Note: These values are consumed by the parent build process
 // They indicate which packages of functionality this OSGi bundle should be delivered inside,
 // or referenced from.
--- a/.github/workflows/cli.yaml
+++ b/.github/workflows/cli.yaml
@@ -36,7 +36,7 @@ jobs:
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v3
         with:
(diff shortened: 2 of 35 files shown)


modules/framework/galasa-parent/dev.galasa.framework.api.launcher/build.gradle
```diff
@@ -9,6 +9,12 @@ dependencies {
     implementation project(':dev.galasa.framework')
 }
 
+test {
+    // Exclude the LauncherTest.java class
+    // As it has no runnable Tests currently, this causes an error in Gradle 9.0.0.
+    exclude '**/LauncherTest.class'
+}
```
eamansour, line 16: Not sure if we already have one open but if not, it would be worth raising a story to re-instate these excluded tests

eamansour (approved):
Looks good!

modules/framework/galasa-parent/dev.galasa.framework.api.launcher/build.gradle
jadecarino: Good idea - I'll open one now.

PR #407, opened 2025-08-29
Downgrading biz.aQute.bnd plugins back down to 6.4.0

jadecarino:
## Why?

PR #405 included changes to uplift the Gradle version used in our builds to 9.0. I upgraded the version of the biz.aQute.bnd plugins to 7.1.0 (latest version) as Gradle 9 removes the Conventions API but the bnd plugins of under version 7 still call it, so this causes errors. Reading the 7.1.0 release notes in more detail, it mentions "All Bnd artifacts are built to run on Java 17 or later." (https://github.com/bndtools/bnd/wiki/Changes-in-7.1.0#backward-compatibility). I am concerned that upgrading to 7.1.0 will break Galasa users who still use Java 11, which our website says we still support.

Therefore I'm downgrading the bnd plugins back to 6.4.0 so still newer than 5.3.0 but does not break backwards compatibility. Also 6.4.0 was used in a few locations so this means a consistent version throughout.

### DIFF
--- a/modules/extensions/galasa-extensions-parent/build.gradle
+++ b/modules/extensions/galasa-extensions-parent/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id 'biz.aQute.bnd.builder' version '7.1.0' apply false
+    id 'biz.aQute.bnd.builder' version '5.3.0' apply false
     id 'dev.galasa.githash' version '0.44.0' apply false
     id 'maven-publish'
     id 'signing'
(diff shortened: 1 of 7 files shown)


jadecarino:
I thought my local build was successful when I downgraded the bnd plugins back to 6.4.0 but they were using a Gradle cache. When I did a local build after clearing my `~/.m2/repository` and `~/.gradle/caches` it was unsuccessful due to the underlying call to the Conventions API, which Gradle 9 removed. To upgrade to Gradle 9 we must use bnd plugins of 7+ but this seems like it breaks Java 11 users.

This needs to be discussed with the team and we may need to revert the Gradle 9 uplift.

PR #408, opened 2025-09-03
Wait for expected text to appear on the terminal screen before making assertions on the JSON content

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2415

## Changes
- Added a `waitForTextInField` call to the failing testToJsonStringReturnsCurrentScreenInJsonFormat method to wait for the expected text to appear on the screen before making assertions on the returned JSON output


techcobweb (approved):
That would do it. Nice one.

PR #424, opened 2025-10-09
Deprecate the callback_url query parameter for GET /auth in favour of base64_callback_url

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2416

The `GET /auth` endpoint previously expected a raw URL to be provided in the `callback_url` query parameter. However, if the supplied URL contained its own query parameters, then the Java servlet code would mistakenly parse those query parameters as if they were intended for the `GET /auth` endpoint, which caused issues.

This PR addresses this problem by deprecating the `callback_url` query parameter in favour of the `base64_callback_url` query parameter, which accepts a Base64 URL encoded URL instead.

## Changes
- Deprecated the `callback_url` query parameter for `GET /auth`
- Added a new `base64_callback_url` query parameter for `GET /auth`
- Updated OpenAPI docs and 0.44.0 release notes


techcobweb (approved):
Looks good.

PR #425, opened 2025-10-10
Only set a test class' result to 'ignored' if all methods were ignored

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2438

## Changes
- When all test methods in a class are ignored, then the test class' result will be set to 'ignored'
- If test methods are allowed to run, then the test class' result will be set to the result of running those test methods
  - If a test method fails, then the test class will be marked as 'failed'
  - If all test methods that ran pass, then the test class will be marked as 'passed'

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
@@ -9,9 +9,11 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Set;
 
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Null;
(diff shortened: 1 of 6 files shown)


modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
```diff
@@ -388,10 +391,21 @@ protected void runTestMethods(@NotNull ITestRunManagers managers, IDynamicStatus
...
                 // method happens inside the testMethod class.
+                testMethodResultsSoFar.add(testMethod.getResult());
                 if (testMethod.fullStop()) {
                     break;
                 }
             }
+
+            // If all of the test methods were ignored, then 'ignored' will be the only result recorded,
+            // so mark the entire test class as ignored as well
+            if (testMethodResultsSoFar.size() == 1) {
```
techcobweb: Even when there are 2 methods ignored ?

techcobweb (approved):

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/TestClassWrapper.java
eamansour: Instead of using the `Result` object, I've updated this to use the string result (e.g. "Failed", "Passed", "Ignored", etc.) in https://github.com/galasa-dev/galasa/pull/425/commits/9827d92d214f9c1ac76bb5175445f91fd2b2b1d6

This set will now collect the unique results that we've encountered when running through the test methods. If multiple methods are ignored, there will only be one 'ignored' result entry in the set.

PR #430, opened 2025-10-23
Allow user-supplied public certificates to be passed into Galasa test pods

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/1825

## Changes
- Updated the code that creates test pod definitions to include a new volume, volume mount, and `JDK_JAVA_OPTIONS` environment variable so that JVMs in test pods can use a cacerts file containing user-supplied public certificates
- Updated release notes and Helm chart installation docs to include instructions on how to supply public certificates into the Helm chart


jadecarino (approved):
Looks good to me

PR #431, opened 2025-10-23
Remove all mention of the old zos3270.terminal.output CPS property since it does not exist

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2449

## Changes
- Removed all remaining mentions of the `zos3270.terminal.output` property as it was removed several releases ago in favour of client-side PNG rendering in the Galasa CLI tool


jadecarino (approved):
Looks good!

PR #432, opened 2025-10-23
Iss2454 - Upgrade grpc-netty-shaded dependency to remove critical vulnerability

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2454

## Changes

- Bumps the `grpc-netty-shaded` dependency to 1.76.0 from 1.68.0 to remove a critical vulnerability reported to us.
- Also bumps the other `io.grpc` group's dependencies in the wrapper and grpc artifact installed in the auth spi to 1.76.0 for consistency.
- Upgrades the version of `protobuf-javalite` in the platform to align with the new `grpc-protobuf-lite` version.

### DIFF
--- a/modules/wrapping/dev.galasa.wrapping.io.grpc.java/pom.xml
+++ b/modules/wrapping/dev.galasa.wrapping.io.grpc.java/pom.xml
@@ -43,7 +43,7 @@
 
 			<!-- 
 			This dependency has a transient dependency of this:
-			com.google.protobuf:protobuf-javalite:jar:3.25.8:compile
+			com.google.protobuf:protobuf-java:jar:3.25.8:compile
 
 			However, we need to depend on it directly or a this causess a NoClassDefFoundError
 			in dev.galasa.framework.auth.spi. We will add an explicit dependency below.
--- a/modules/wrapping/dev.galasa.wrapping.io.grpc.java/pom.xml
+++ b/modules/wrapping/dev.galasa.wrapping.io.grpc.java/pom.xml
@@ -40,25 +40,11 @@
 			<groupId>io.grpc</groupId>
 			<artifactId>grpc-protobuf</artifactId>
 			<version>1.76.0</version>
-
-			<!-- 
-			This dependency has a transient dependency of this:
-			com.google.protobuf:protobuf-java:jar:3.25.8:compile
-
-			However, we need to depend on it directly or a this causess a NoClassDefFoundError
-			in dev.galasa.framework.auth.spi. We will add an explicit dependency below.
-			-->
-			<exclusions>
-				<exclusion>
-					<groupId>com.google.protobuf</groupId>
-					<artifactId>protobuf-java</artifactId>
-				</exclusion>
-			</exclusions>
 		</dependency>
 
 		<!--
 		This dependency is relied upon by io.grpc:grpc-protobuf
-		but we want to explicitly depend on it here.
+		but we need to depend on it directly to avoid OSGi issues.
 		-->
 		<dependency>
 			<groupId>com.google.protobuf</groupId>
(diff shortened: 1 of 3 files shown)


modules/wrapping/dev.galasa.wrapping.io.grpc.java/pom.xml
```diff
@@ -19,71 +19,71 @@
...
 		<dependency>
 			<groupId>io.grpc</groupId>
 			<artifactId>grpc-protobuf</artifactId>
-			<version>1.68.0</version>
+			<version>1.76.0</version>
 
 			<!-- 
 			This dependency has a transient dependency of this:
-			com.google.protobuf:protobuf-java:jar:3.25.3:compile
+			com.google.protobuf:protobuf-javalite:jar:3.25.8:compile
```
eamansour: Minor - this comment has been changed to `protobuf-javalite`, should it be `protobuf-java`?

modules/wrapping/dev.galasa.wrapping.io.grpc.java/pom.xml
```diff
@@ -19,71 +19,71 @@
...
-			a later version.
+			This causes a NoClassDefFoundError in dev.galasa.framework.auth.spi.
+			We will add an explicit dependency on a later version that fixes this.
 			-->
 			<exclusions>
 				<exclusion>
-				<groupId>com.google.protobuf</groupId>
-				<artifactId>protobuf-javalite</artifactId>
+					<groupId>com.google.protobuf</groupId>
+					<artifactId>protobuf-javalite</artifactId>
```
eamansour: Same here - I don't think we need this exclusion anymore. Looking at https://mvnrepository.com/artifact/io.grpc/grpc-protobuf-lite/1.76.0 - the protobuf-javalite dependency ([3.25.8](https://mvnrepository.com/artifact/com.google.protobuf/protobuf-javalite/3.25.8)) doesn't have any vulnerabilities, so we could just use that as-is.

This exclusion was added because the 3.25.3 version was vulnerable so we wanted to upgrade it to a version that isn't vulnerable.

modules/wrapping/dev.galasa.wrapping.io.grpc.java/pom.xml
```diff
@@ -19,71 +19,71 @@
...
-			a later version.
+			However, we need to depend on it directly or a this causess a NoClassDefFoundError
+			in dev.galasa.framework.auth.spi. We will add an explicit dependency below.
 			-->
 			<exclusions>
 				<exclusion>
-				<groupId>com.google.protobuf</groupId>
-				<artifactId>protobuf-java</artifactId>
+					<groupId>com.google.protobuf</groupId>
+					<artifactId>protobuf-java</artifactId>
```
eamansour: I don't think this exclusion is actually needed anymore. Looking at https://mvnrepository.com/artifact/io.grpc/grpc-protobuf/1.76.0 - the 3.25.8 version of protobuf-java isn't vulnerable, so we could use it as-is.

This was originally added in because the 3.25.3 version was vulnerable.

modules/wrapping/dev.galasa.wrapping.io.grpc.java/pom.xml
jadecarino: Thanks, I've reverted this back in https://github.com/galasa-dev/galasa/pull/432/commits/f1b4b034dab22e007195b674d8019e17fea4e298

modules/wrapping/dev.galasa.wrapping.io.grpc.java/pom.xml
jadecarino: Thanks Eamonn, as discussed on Slack I've now removed the exclusions but kept the direct dependencies in. I've made this change in https://github.com/galasa-dev/galasa/pull/432/commits/71ec7146e0222fbcdcbb07d852dbdefa449ace79

modules/wrapping/dev.galasa.wrapping.io.grpc.java/pom.xml
jadecarino: Thanks Eamonn, as discussed on Slack I've now removed the exclusions but kept the direct dependencies in. I've made this change in https://github.com/galasa-dev/galasa/commit/71ec7146e0222fbcdcbb07d852dbdefa449ace79

modules/wrapping/dev.galasa.wrapping.io.grpc.java/pom.xml
```diff
@@ -19,94 +19,63 @@
...
 		<!--
 		This dependency is relied upon by io.grpc:grpc-protobuf
-		but we want to upgrade a transient dependency there.
+		but we need to depend on it directly to avoid OSGi issues.
 		-->
 		<dependency>
 			<groupId>com.google.protobuf</groupId>
 			<artifactId>protobuf-java</artifactId>
-			<version>3.25.5</version>
+			<version>3.25.8</version>
```
jadecarino, line 52: There is still no version for this in the platform hence why its explicit but I'll create a new story to align all versions of this in the project and then update this.

eamansour (approved):

PR #433, opened 2025-10-27
Iss2406 - Simplatform documentation update

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2406

## Changes

- Update instructions in the documentation about cloning the Simplatform repo, for users to clone the latest released tag, rather than the `main` development branch. Not only could the main development branch be unstable as its still in development, but also using the released version of galasactl with a newer Simplatform OBR causes OSGi issues.
- As these lines in the docs now refer to the released version, they'll need to be kept up to date, so these have been added to the set-version.sh script.


eamansour (approved):
Looks good, thanks for making these changes!

PR #434, opened 2025-10-28
Updated Cemt Manager

ayushikumari3:
This PR extends the CEMT Manager interface to provide more granular control and verification capabilities for CICS resource management through CEMT commands.
New APIs enable direct inquiry, state checking, and waiting mechanisms for CICS resources during automation or runtime operations.

Key Changes:- 

🆕 Added: inquireResource(...) (overloaded)
  New overload returns a boolean result based on whether the provided searchText is found in the inquiry output.
 
🆕 Added: inquireResourceNotFound(...) 
Performs a CEMT inquiry and returns true if the resource is not found.

🆕 Added: isResourceEnabled(...)
Executes a CEMT inquiry and checks if the target resource is in the ENABLED state.

🆕 Added: waitForDisabledResource(...) 
Adds a no-timeout variant of the existing method.
Waits until a resource becomes disabled by detecting "Dis" in inquiry results.

🆕 Added: waitForEnabledResource(...)
Adds a no-timeout variant that waits for a resource to become enabled by detecting "Ena" in inquiry results.


modules/managers/galasa-managers-parent/galasa-managers-cicsts-parent/dev.galasa.cicsts.cemt.manager/src/main/java/dev/galasa/cicsts/cemt/internal/CemtImpl.java
```diff
@@ -5,24 +5,40 @@
...
 
 public class CemtImpl implements ICemt {
 
+    private static final Log logger = LogFactory.getLog(CemtImpl.class);
+    
     private ICicsRegion cicsRegion;
 
-    public CemtImpl(ICicsRegion cicsRegion) {
+    /** Holds the defaultResourceTimeout. */
+    private int defaultResourceTimeout;
```
eamansour: It would be good if we could also include the time unit that this resource timeout is expected to be in (e.g. milliseconds, seconds, etc.) to save us from having to look it up from the CPS property.

If this is in milliseconds, a `long` value would be more suitable than an `int`.

Also, another minor comment: the Javadoc for this field isn't very useful and could be removed

So for example, the field declaration could become:

```suggestion
    private long defaultResourceTimeoutMilliseconds;
```

modules/managers/galasa-managers-parent/galasa-managers-cicsts-parent/dev.galasa.cicsts.cemt.manager/src/main/java/dev/galasa/cicsts/cemt/internal/CemtImpl.java
```diff
@@ -5,24 +5,40 @@
...
+    
+    public CemtImpl(ICicsRegion cicsRegion) throws CemtManagerException {
         this.cicsRegion = cicsRegion;
+        this.defaultResourceTimeout = DefaultResourceTimeout.get(cicsRegion.getZosImage())*1000;
```
eamansour: It's not immediately obvious what time unit `DefaultResourceTimeout` is expected to be in. From the looks of this, it looks like we're converting seconds to milliseconds, so maybe `DefaultResourceTimeout` should be:

```suggestion
        this.defaultResourceTimeoutMilliseconds = DefaultResourceTimeoutSeconds.get(cicsRegion.getZosImage())*1000;
```

modules/managers/galasa-managers-parent/galasa-managers-cicsts-parent/dev.galasa.cicsts.cemt.manager/src/main/java/dev/galasa/cicsts/cemt/internal/CemtImpl.java
```diff
@@ -366,5 +449,164 @@ public boolean performSystemProperty(@NotNull ICicsTerminal terminal, @NotNull S
...
+        try {
+            long timeoutTimeInMilliseconds = Calendar.getInstance()
+                .getTimeInMillis() + resourceTimeout;
+            while (Calendar.getInstance().getTimeInMillis() < timeoutTimeInMilliseconds) {
+                terminal.resetAndClear();
+                terminal.type("CEMT INQUIRE " + resourceType + "(" + resourceName + ")").enter().waitForKeyboard();
+                if (terminal.retrieveScreen().contains(" Dis ")) {
+                    return;
+                } 
+                if (Calendar.getInstance().getTimeInMillis() >= timeoutTimeInMilliseconds) {
```
eamansour: The `Calendar.getInstance().getTimeInMillis()` is slightly slower than doing a `Instant.now()` or `System.currentTimeMillis()` due to the need to get a `Calendar` object, so maybe we could use one of those instead?

modules/managers/galasa-managers-parent/galasa-managers-cicsts-parent/dev.galasa.cicsts.cemt.manager/src/main/java/dev/galasa/cicsts/cemt/internal/CemtImpl.java
```diff
@@ -366,5 +449,164 @@ public boolean performSystemProperty(@NotNull ICicsTerminal terminal, @NotNull S
...
+
+    @Override
+    public void waitForEnabledResource(ICicsTerminal terminal, String resourceType, String resourceName,
+        int resourceTimeout) throws CemtException
+    {
+        // Calculate when we should timeout
+        long timeoutTimeInMilliseconds = Calendar.getInstance().getTimeInMillis() + resourceTimeout;
+
+        // Keep going until we reach timeout time
+        while (Calendar.getInstance().getTimeInMillis() < timeoutTimeInMilliseconds) {
```
eamansour: Same comment here:

The `Calendar.getInstance().getTimeInMillis()` is slightly slower than doing a `Instant.now()` or `System.currentTimeMillis()` due to the need to get a `Calendar` object, so maybe we could use one of those instead?

modules/managers/galasa-managers-parent/galasa-managers-cicsts-parent/dev.galasa.cicsts.cemt.manager/src/main/java/dev/galasa/cicsts/cemt/internal/CemtImpl.java
```diff
@@ -366,5 +449,164 @@ public boolean performSystemProperty(@NotNull ICicsTerminal terminal, @NotNull S
...
+            terminal.type("CEMT INQUIRE " + resourceType + "(" + resourceName + ")").enter().waitForKeyboard();
+            
+            if (terminal.retrieveScreen().contains(" Ena "))
+            {
+                return;
+            }
+
+            // Double check the timeout time, just in case the the send text
+            // took ages
+            if (Calendar.getInstance().getTimeInMillis() >= timeoutTimeInMilliseconds)
```
eamansour: Same comment here:

The `Calendar.getInstance().getTimeInMillis()` is slightly slower than doing a `Instant.now()` or `System.currentTimeMillis()` due to the need to get a `Calendar` object, so maybe we could use one of those instead?

modules/managers/galasa-managers-parent/galasa-managers-cicsts-parent/dev.galasa.cicsts.manager/src/main/java/dev/galasa/cicsts/ICemt.java
```diff
@@ -31,6 +31,69 @@ public CicstsHashMap inquireResource(@NotNull ICicsTerminal cemtTerminal,
...
+     * @param cemtTerminal an {@link ITerminal} object logged on to the CICS region and in an active CEMT session.
+     * If mixed case is required, the terminal should be presented with no upper case translate status.
+     * @param resourceType a {@link String} of the resource type you are looking for.
+     * @param resourceName a {@link String} of the name of the resource you are looking for.
+     * @param resourceTimeout
+     * @throws CemtException
+     */
+    public void waitForDisabledResource(ICicsTerminal terminal,
+            String resourceType, String resourceName,
+            int resourceTimeout) throws CemtException;
```
eamansour: Similar issue here regarding timeout - it's not clear what time unit `int resourceTimeout` parameter should be given in, and it isn't mentioned in the javadoc.

modules/managers/galasa-managers-parent/galasa-managers-cicsts-parent/dev.galasa.cicsts.manager/src/main/java/dev/galasa/cicsts/ICemt.java
```diff
@@ -31,6 +31,69 @@ public CicstsHashMap inquireResource(@NotNull ICicsTerminal cemtTerminal,
...
+     * @param cemtTerminal an {@link ITerminal} object logged on to the CICS region and in an active CEMT session.
+     * If mixed case is required, the terminal should be presented with no upper case translate status.
+     * @param resourceType a {@link String} of the resource type you are looking for.
+     * @param resourceName a {@link String} of the name of the resource you are looking for.
+     * @param resourceTimeout
+     *            - The timeout in milliseconds.
+     * @throws CEMTException
+     */
+    public void waitForEnabledResource(ICicsTerminal terminal, String resourceType,
+        String resourceName, int resourceTimeout) throws CemtException;
```
eamansour: The Javadoc for the `resourceTimeout` mentions that it is a timeout in milliseconds. In this case, I think a `long` would be more suitable for millisecond values and it would be great if we could rename the `resourceTimeout` parameter to something like:

```java
long resourceTimeoutInMilliseconds
```

eamansour (commented):
Hi @ayushikumari3, thanks for submitting this PR!

Overall the changes look good - I've added a few comments around clarifying the time unit for the default resource timeout value and a few around something like `Instant.now()` instead of using the old `Calendar` class. These are all minor issues that would be great to sort out whenever you get the chance to do so.

Let me know if anything doesn't make sense!

modules/managers/galasa-managers-parent/galasa-managers-cicsts-parent/dev.galasa.cicsts.cemt.manager/src/main/java/dev/galasa/cicsts/cemt/internal/CemtImpl.java
ayushikumari3: Updated code as per suggestion

modules/managers/galasa-managers-parent/galasa-managers-cicsts-parent/dev.galasa.cicsts.cemt.manager/src/main/java/dev/galasa/cicsts/cemt/internal/CemtImpl.java
ayushikumari3: Updated code as per suggestion

modules/managers/galasa-managers-parent/galasa-managers-cicsts-parent/dev.galasa.cicsts.manager/src/main/java/dev/galasa/cicsts/ICemt.java
ayushikumari3: Updated javadoc and renamed the variable

modules/managers/galasa-managers-parent/galasa-managers-cicsts-parent/dev.galasa.cicsts.manager/src/main/java/dev/galasa/cicsts/ICemt.java
ayushikumari3: Renamed the variable

modules/managers/galasa-managers-parent/galasa-managers-cicsts-parent/dev.galasa.cicsts.cemt.manager/src/main/java/dev/galasa/cicsts/cemt/internal/CemtImpl.java
eamansour: The javadoc for `DefaultResourceTimeout` already mentions that the timeout is given in seconds. Happy to leave this as-is.

eamansour (commented):
Looks good, thanks for making those changes!

We're currently in the process of releasing Galasa 0.44.0, so a change freeze is in effect. I'll approve and merge this PR into main once 0.44.0 has been released.

eamansour (approved):

PR #435, opened 2025-10-28
Remove the VTP manager as ZVTP has been withdrawn from the market

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2460

See https://www.ibm.com/docs/en/announcements/software-support-discontinuance-z-platform-select-programs-some-replacements-available?region=US

## Changes
- Removed the VTP manager
- Removed all docs about the VTP manager now that it no longer exists


jadecarino (approved):
Looks good, thanks!

PR #439, opened 2025-10-31
Clean up queued local runs from the DSS after a period of time

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2395

The logs and DSS state attached to the above issue indicate that the resource monitor was not cleaning up a 'local' run that was stored in the DSS (in this case, it seems like a shared DSS was being used for remote and local runs).

As the run was 'queued', it didn't have a heartbeat, so the dead heartbeat monitor wasn't picking it up, and because it was a local run, the engine controller was ignoring it (correctly). This meant that nothing was looking to clean it up if the user stopped the run before it progressed to a different state.

This PR adds a check for local runs that have been in the DSS for too long, and marks them as 'Hung' if they exceed a given timeout value, which defaults to 15mins but can be customised using a `framework.resource.management.local.queued.timeout` CPS property.

## Changes
- Renamed the `RunAllocatedRunCleanup` job in the resource monitor to `RunInactiveRunCleanup`
- Added logic to clean up local runs in the 'queued' state that have been in the DSS for too long


modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/RunInactiveRunCleanup.java
```diff
@@ -0,0 +1,124 @@
...
+            }
+        }
+    }
+
+    /**
+     * Gets the maximum number of seconds that local runs can remain in the 'queued' state before needing to
+     * be cleaned up.
+     *
+     * This value can be customised using the CPS property:
+     *     framework.resource.management.queued.timeout=[number of seconds]
```
jadecarino: Minor - this is missing a `local` in there

jadecarino (changes requested):
One minor change but otherwise looks great!

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/RunInactiveRunCleanup.java
eamansour: Great spot, thanks! Fixed in https://github.com/galasa-dev/galasa/pull/439/commits/100a95b24e4c974d41c7f490872b150bd164c8b9

jadecarino (approved):

PR #444, opened 2025-11-07
James c dev container

James-Cocker:
Adds a dev container to Galasa


.devcontainer/dev-container-setup.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# Sets envs if they are specified on local machine (prevents null values)
+function set_env_vars() {
+  local GALASA_TOKEN_VALUE="$1"
+  local GALASA_API_SERVER_URL_VALUE="$2"
+  local SOURCE_MAVEN_VALUE="$3"
+  local GALASA_BOOTSTRAP_VALUE="$4"
+  local GPG_PASSPHRASE_VALUE="$5"
+
+  local PROFILE_FILE="/home/node/.bashrc"
```
eamansour: use `${HOME}` instead of a hard-coded user folder? I doubt the `node` user would exist if you're not using a Node.js image.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,57 @@
+// /*
+//  * Copyright contributors to the Galasa project
+//  *
+//  * SPDX-License-Identifier: EPL-2.0
+//  */
```
eamansour: Very minor - double comments here, this could just be:

```
/*
 * Copyright contributors to the Galasa project
 *
 * SPDX-License-Identifier: EPL-2.0
 */
```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,57 @@
...
+      "moby": false
+		}
+  },
+
+  "mounts": [
+    "source=${localEnv:HOME}/.m2,target=/home/root/.m2,type=bind,consistency=cached",
+    "source=/var/run/docker.sock,target=/var/run/docker-host.sock,type=bind"
+  ],
+
+  "remoteUser": "root",
```
eamansour: Is there a reason why the dev container is running as root?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,57 @@
...
+			"version": "3.11.0"
+    },
+    
+    "ghcr.io/devcontainers/features/go:1": {
+			"version": "1.23.5"
+    },
+    
+    "ghcr.io/devcontainers/features/node:1": {
+      "version": "20.10.0"
+    },
```
eamansour: Do we need to install node? The Galasa repo doesn't use Node.js anywhere

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,57 @@
...
+
+    "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {
+			"version": "28.3.3-1",
+      "moby": false
+		}
+  },
```
eamansour: If someone doesn't have docker installed outside of the dev container, would the docker-outside-of-docker feature cause issues?

.github/workflows/maven.yaml
```diff
@@ -80,8 +80,10 @@ jobs:
           chmod '700' /home/runner/work/gpg
 
       - name: Import GPG
+        env:
+          GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
         run: |
-          gpg --homedir /home/runner/work/gpg --pinentry-mode loopback --passphrase-file /home/runner/work/secrets/passphrase.file --import /home/runner/work/secrets/galasa.gpg
+          gpg --homedir /home/runner/work/gpg --passphrase $GPG_PASSPHRASE --import /home/runner/work/secrets/galasa.gpg
```
eamansour: Not 100% sure about this - would this print the `GPG_PASSPHRASE` out into the workflow logs?

Have you tried running a main build workflow in a fork to check this?

Edit: it won't print the passphrase out since GH actions knows the value comes from the repo's secrets

eamansour: Given that this PR is focusing on adding a dev container, I don't think it should be touching the GitHub Actions workflows at all

modules/maven/galasa-maven-plugin/pom.xml
```diff
@@ -219,10 +219,7 @@
 							<goal>sign</goal>
 						</goals>
 						<configuration>
-							<gpgArguments>
-								<arg>--pinentry-mode</arg>
-								<arg>loopback</arg>
-							</gpgArguments>
+							<passphrase>${env.GPG_PASSPHRASE}</passphrase>
```
eamansour: Not 100% sure about this change - if someone doesn't have GPG_PASSPHRASE in their environment, then this would fail to sign artifacts.

The loopback I believe would have at least prompted the user for a passphrase.

Is this change really needed?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,57 @@
...
+			"version": "1.23.5"
+    },
+    
+    "ghcr.io/devcontainers/features/node:1": {
+      "version": "20.10.0"
+    },
+
+    "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {
+			"version": "28.3.3-1",
+      "moby": false
```
eamansour, line 31: Should `moby` be set to true so that it uses the Docker API? Is there a reason why it's set to false?

eamansour (commented):
Looks good overall - just a few initial comments and questions

Note: this PR links to https://github.com/galasa-dev/projectmanagement/issues/2349

modules/maven/galasa-maven-plugin/pom.xml
James-Cocker: I was talking this over with Jade and Mike and we came to the conclusion that everything should be automated, and given that other parts of the build script require the variable anyways it made sense. The main reason that prompted this change is that as the GPG agent is running in restricted mode in the dev container, and it wasn't very happy with pinentry-mode.

.github/workflows/maven.yaml
James-Cocker: My apologies, I went ahead and changed all mentions of pinentry mode but i can reverse this if you'd like. Although, I do wonder why github actions would need script interaction with pinentry mode if it's meant to be an automated workflow.

.devcontainer/devcontainer.json
James-Cocker: I'm not too sure, I installed node as it it's required in the install guide. Should i remove Node from the install guide as well? https://github.com/galasa-dev/galasa/blob/main/developer-docs/install-pre-req-tools.md

.devcontainer/dev-container-setup.sh
James-Cocker: This was my apologies, I noticed this error as soon as i popped you that message on slack, along with some errors in the variable forwarding so this has since changed but I'm happy to implement the $HOME variable depending on your answer about the root user

.devcontainer/devcontainer.json
James-Cocker: This was one of my earlier changes so I'll run it again as the default `vscode` user - but I believe there were a lot of permission denied errors which was my reasoning. Can damage be caused if run as root?

.devcontainer/devcontainer.json
James-Cocker: I believe there was an error that occurred with it installing the OSS Moby Build instead of Docker CE (https://github.com/devcontainers/features/tree/main/src/docker-outside-of-docker), but I must admit i didn't attempt to spin up the container with moby: false so I'll test that works and get back to you on this.

.devcontainer/devcontainer.json
eamansour: Yes, running containers as root carries several security risks.

If you need to set things up as root, you would typically do so by making a Dockerfile that sets up things it needs as the root user, then switches to a standard user for containers to then use.

.devcontainer/devcontainer.json
eamansour: Node.js isn't used in the main Galasa repo, so yes I would say it's not needed in those docs and isn't needed in the dev container. We only use Node.js in the webui, and some people might only be interested in making changes to managers for local test runs, so won't need to set up the webui at all.

.github/workflows/maven.yaml
eamansour: I'm not sure why pinentry mode is used in here either, but the workflow passes in a passphrase-file so that must be why it doesn't prompt for a passphrase in the workflow.

I'm just wondering if there's a way for the dev container to keep using the same settings without having to change the build process.

If not, then I think this change might be fine - it's worth making sure that Jade is aware of this as well and we likely won't need this step in the workflow above if we're using `--passphrase` https://github.com/galasa-dev/galasa/pull/444/files/51fc6dd47bfebc3909d45226dff9e72cf0cff106#diff-edf39caa9c276bfb5df0ef3ef090f9e0577ce28316ef8a6ff3ef7eb5b155c21dL57-L61

James-Cocker: Agreed, not having to change the build process would be ideal! Once I've sorted out the root user I'll do another brief search around and then forward over the error to you if you're keen to stop the build process changing.

James-Cocker: Yes no worries I believe Jade is already aware as she and Mike helped me make the decision. When you mention `we likely won't need this step in the workflow` did you mean you would like me to leave the change or to reverse it?

eamansour: > Yes no worries I believe Jade is already aware as she and Mike helped me make the decision. When you mention `we likely won't need this step in the workflow` did you mean you would like me to leave the change or to reverse it?

If you find a way that works without having to change the build process, that would be great.

If not, then your change can either be kept as-is to use `--passphrase` or you could use the `--passphrase-file` as it was previously. The step here https://github.com/galasa-dev/galasa/pull/444/files/51fc6dd47bfebc3909d45226dff9e72cf0cff106#diff-edf39caa9c276bfb5df0ef3ef090f9e0577ce28316ef8a6ff3ef7eb5b155c21dL57-L61 creates a file containing the GPG passphrase, which won't be used if we switch to using `--passphrase`, so that step can be removed if it's not being used anywhere else in the workflow.

It would be good if you could test out the workflow on a fork by following the docs about setting up a fork in the [CONTRIBUTING.md](https://github.com/galasa-dev/galasa/blob/main/CONTRIBUTING.md) file.

.devcontainer/devcontainer.json
eamansour: Ahh okay no worries if it causes issues, happy for you to keep this as-is - I was just curious :)

.devcontainer/dev-container-setup.sh
James-Cocker: Update: done

.devcontainer/devcontainer.json
James-Cocker: Update: solved, turns out it only denied to run the dev-container-setup.sh, so I simply used chmod

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,53 @@
...
+		}
+  },
+
+  "mounts": [
+    "source=${localEnv:HOME}/.m2,target=/home/root/.m2,type=bind,consistency=cached",
```
eamansour: Should this mount target be pointing to a user other than `/home/root`?

James-Cocker: Most definitely - good spot

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,53 @@
...
+	"image": "mcr.microsoft.com/devcontainers/java:17",
+  
+  "features": {
+    "ghcr.io/devcontainers/features/java:1": {
+      "jdkDistro": "sem",
+      "version": "17.0.14-sem",
+			"installMaven": "true",
+			"mavenVersion": "3.9.0",
+			"installGradle": "true",
+			"gradleVersion": "8.14.3"
```
eamansour: Can the gradle version be set to Gradle 9.0.0 like in the GH Actions workflows? I think it would be good to keep the dev container consistent with the version we use in the workflows

modules/maven/galasa-maven-plugin/pom.xml
```diff
@@ -222,7 +222,7 @@
 							<gpgArguments>
 								<arg>--pinentry-mode</arg>
 								<arg>loopback</arg>
-							</gpgArguments>
+							</gpgArguments>
```
eamansour: Really minor - can this whitespace be trimmed?

eamansour (approved):

PR #445, opened 2025-11-12
Add method to get CPS properties and overrides for an existing test run

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2476

## Changes
- Added a new `getCpsPropertiesAndOverridesUsedByTestRun` method to the IFrameworkRuns interface to retrieve CPS properties from a list of namespaces and any overrides used by a test run with the given run name.
  - This can be used within IResourceManagementProvider services to clean up resources using the CPS and overrides provided to a test run.


modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/IFrameworkRuns.java
```diff
@@ -63,4 +64,20 @@ IRun submitRun(String type, String requestor, String bundleName, String testName
     boolean markRunCancelling(String runName, TestRunLifecycleStatus currentStatus) throws DynamicStatusStoreException ;
 
     void addRunRasAction(IRun run, RunRasAction rasActionToAdd) throws DynamicStatusStoreException;
+
+    /**
+     * Returns a map of CPS properties and overrides that were used by a test run with the given run name.
+     * 
+     * The returned map combines the current CPS properties from the given namespaces with the overrides
```
jadecarino, line 71: Do we see an issue with this if a CPS property or set of them are edited in between the test run and the resource management process? I suppose there's not another option given that we don't store all the CPS properties used in the DSS.

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/IFrameworkRuns.java
```diff
@@ -63,4 +64,20 @@ IRun submitRun(String type, String requestor, String bundleName, String testName
...
+     * associated with the given test run, so any overrides that were used will take precedence.
+     * 
+     * If no test run with the given run name exists, then the current CPS properties from the provided
```
jadecarino, line 74: Just to clarify I understand, is this in case the DSS properties for that run have already been cleared up, or if the run doesn't/never existed?  Would this method ever be called for a run that never existed?

jadecarino (approved):
Changes look good so approved, I just have a couple of questions around the logic to ensure I understand

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/IFrameworkRuns.java
eamansour: That's right - I don't believe this would happen as there's a configurable period of time between a run being marked as "finished" and then being removed. That time window exists so that services, like the resource management process, can pick up the run from the DSS and process it before it gets cleaned out.

That said, there's always a chance that someone configures the time window to be too short, so this acts as a fallback if that happens.

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/IFrameworkRuns.java
eamansour: Yes, there's a chance that this could happen. The test run does save away the CPS properties that the test accessed but that happens at the very end of the test run, which may not be invoked if the run exited towards the start.

So yes, there's no reliable way at the moment to get a snapshot of the CPS properties when the test run was submitted unfortunately.

eamansour: I'm now wondering if we should store the record of accessed CPS properties for a test run in the DSS record for that run. At the moment, the accessed CPS properties are only stored in-memory by the framework inside the test run's pod.

A few downsides to this would be:
- Test runs would write to the DSS a lot more frequently
- Potentially storing a lot more data in the DSS if a test run accesses a lot of CPS properties
- Incomplete CPS property access records if the test run exited before accessing the CPS properties it needs (might not be an issue, but still worth noting)

PR #447, opened 2025-11-13
Upgrade SSH client library used in the IP network manager

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2461

## Changes
These changes have been tested out by running both local tests and remote tests (including the ZosManagerTSOCommandIVT test) on a Galasa service installed on minikube that use the IP network manager to connect to various servers via SSH. No connection issues or OSGi-related issues have been encountered.

- Replace the com.jcraft:jsch library with the actively maintained fork com.github.mwiede:jsch to support more modern algorithms
  - Note: this drops support for the insecure RSA/SHA1 algorithm
- Removed the dev.galasa:dev.galasa.wrapping.com.jcraft.jsch bundle as the forked library is available as an OSGi bundle
- Added a release note to inform users of this change and provide advice on upgrading their servers if they are using RSA/SHA1


jadecarino (approved):
Changes look good to me - thanks Eamonn

PR #452, opened 2025-11-19
galasa-maven-plugin to handle target/classes dependencies

ideeley:
Fix for https://github.com/galasa-dev/projectmanagement/issues/2483

Sometimes during a test project build, e.g. when javadoc is being generated, Maven generates dependency artifacts for the project before the bundle is built - these artifacts therefore point to the target/classes folder instead of the jar file. These artifacts are then passed to the galasa-maven-plugin, which doesn't know what to do with them....

This change modifies BuildOBRResources to look for the jar file in the parent folder if the artifact is pointing to a "classes" folder, as the jar _should_ exist by this time. If it's not found, we just ignore the dependency artifact, so we're no worse off than before.


jadecarino (approved):
Thanks for contributing this fix Ian!

PR #459, opened 2025-11-28
Add ability to load and run resource management providers once

eamansour:
## Why?
Part 1 of https://github.com/galasa-dev/projectmanagement/issues/2484

## Changes
- Added a new `--local-resource-managment` flag to the boot jar which runs a version of resource management that runs the loaded IResourceManagementProviders only once
  - Refactored common code between the existing ResourceManagement class and the new LocalResourceManagement class to avoid code duplication

- Added a new `runOnce()` method to the IResourceManagementProvider interface which can optionally be implemented to allow a user to run the resource management provider once, rather than scheduling a job to run indefinitely

- Added new `--includes-monitor-pattern` and `--excludes-monitor-pattern` flags that can be supplied multiple times with glob patterns representing the resource monitors that should be included or excluded as part of the resource management run
  - The existing `GALASA_MONITOR_INCLUDES_GLOB_PATTERNS` and `GALASA_MONITOR_EXCLUDES_GLOB_PATTERNS` environment variables can still be used to supply glob patterns


modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/init/ResourceManagerInitStrategy.java
```diff
@@ -5,15 +5,5 @@
  */
 package dev.galasa.framework.internal.init;
 
-import dev.galasa.framework.Framework;
-import dev.galasa.framework.spi.FrameworkException;
-import dev.galasa.framework.spi.IConfigurationPropertyStoreService;
-
 public class ResourceManagerInitStrategy extends DefaultInitStrategy {
```
jadecarino, line 8: Why do we no longer need this method? Assuming we need to keep this class even though it no longer does anything just to not break any other classes that return it?

modules/framework/galasa-parent/dev.galasa.framework/src/test/java/dev/galasa/framework/internal/init/TestResourceManagerInitStrategy.java
```diff
@@ -5,35 +5,17 @@
...
     
 import dev.galasa.framework.GalasaFactory;
 import dev.galasa.framework.IFrameworkInitialisationStrategy;
-import dev.galasa.framework.mocks.MockFramework;
-
 
 public class TestResourceManagerInitStrategy {
 
     @Test
     public void testCanCreateStrategy() {
```
jadecarino, line 16: I see, okay :)

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/init/ResourceManagerInitStrategy.java
jadecarino: Discussed on call.

jadecarino (approved):

PR #468, opened 2025-12-09
Include user priority field in test run scheduling priority calculation

eamansour:
## Why?

This PR makes the required changes to include the `priority` field in user records when calculating the priority for queued test runs. Note: this priority field is not made settable in this PR, so it will default to 0 for now.

## Changes
- Updated the PrioritySchedulingService code to get the priority value for a queued run's requestor and add that value to the priority total for that test run.
  - When getting user priorities, the retrieved priorities are cached in the DSS to avoid having to reach out to CouchDB frequently
- Fixed an issue where the engine controller was not able to initialise the auth store as part of its framework initialisation.
  - `initialiseAuthStore` has been added to the IFrameworkInitialisation interface. Its implementation has moved from the ApiServerInitialisation class (which is no longer needed now as it was only used to initialise the auth store) into the FrameworkInitialisation class itself


modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStore.java
```diff
@@ -127,24 +127,53 @@ public void put(@NotNull String key, @NotNull String value) throws DynamicStatus
...
+
+    /**
+     * Put multiple key-value pairs into the DSS that expire after a given amount of time in seconds
+     * 
+     * @param keyValues the key-value pairs to be stored
+     * @param timeToLiveSecs the amount of time in seconds for the key-value pairs to be available for before expiring
+     * @throws DynamicStatusStoreException if there was an error accessing etcd
+     */
+    @Override
+    public void put(@NotNull Map<String, String> keyValues, @NotNull long timeToLiveSecs)
```
eamansour, line 142: Previously, we only had the ability to set a single property with a time-to-live. Since we need to set multiple properties into the DSS to cache a user, this has been added to allow us to set multiple properties at once with the same time-to-live.

modules/framework/galasa-parent/dev.galasa.framework.api/src/main/java/dev/galasa/framework/api/internal/ApiServerInitialisation.java
eamansour, line 1: This is no longer needed as the code to initialise the auth store has been lifted up into the FrameworkInitialisation class.

jadecarino:
Just a note on this as it's slightly related to the story I am currently working on (https://github.com/galasa-dev/projectmanagement/issues/2389) - when my changes are delivered, we may need to update this to check both the queued run's requestor and user, as they may be different. Requestor is to be the username related to the JWT and user is the person who actually kicked it off - this is for automation purposes.

modules/framework/galasa-parent/dev.galasa.framework.api/src/main/java/dev/galasa/framework/api/internal/ApiServerInitialisation.java
jadecarino: Okay good to know thanks!

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStore.java
jadecarino: Okay great, thanks

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/Framework.java
```diff
@@ -69,6 +71,7 @@ public class Framework implements IFramework, IShuttableFramework {
     private IAuthStore                         authStore;
     private RBACService                        rbacService;
     private IStreamsService                    streamsService;
+    private CacheUsers                         usersCache;
```
jadecarino: Is this needed? It looks like it's not doing anything

modules/framework/galasa-parent/dev.galasa.framework/src/test/java/dev/galasa/framework/internal/rbac/TestCacheRBACImpl.java
```diff
@@ -107,12 +106,9 @@ public void testIsActionPermittedReturnsFalseForInvalidMappings() throws Excepti
         mockAuthStoreService.addUser(mockUser);
         MockRBACService mockRbacService = FilledMockRBACService.createTestRBACServiceWithTestUser(loginId);
 
-        CacheRBAC cache = new CacheRBACImpl(mockDssService, mockAuthStoreService, mockRbacService);
-
-        List<Action> permittedActions = List.of(GENERAL_API_ACCESS.getAction(), SECRETS_GET_UNREDACTED_VALUES.getAction());
-        Set<String> permittedActionsIds = permittedActions.stream().map(Action::getId).collect(Collectors.toSet());
```
jadecarino, line 113: Assuming you have removed this because it was redundant as these are the Admin role's actions?

jadecarino: Or because they are not used in the test case?

jadecarino (approved):

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/Framework.java
eamansour: Great spot, thanks - removed in https://github.com/galasa-dev/galasa/pull/468/commits/019cb96c7d09d031ed4a92ca63674dcc1fd13a29

modules/framework/galasa-parent/dev.galasa.framework/src/test/java/dev/galasa/framework/internal/rbac/TestCacheRBACImpl.java
eamansour: That's right - those actions weren't needed because the actions are already set into the role that the mock RBAC service creates when calling `FilledMockRBACService.createTestRBACServiceWithTestUser(loginId)`.

However, to make the unit test easier to understand and to avoid having to look inside the mock RBAC service, I've added the permitted actions back in to override the default role that's created by the mock RBAC service using `FilledMockRBACService.createTestRBACServiceWithTestUser(loginId, permittedActions)` - see https://github.com/galasa-dev/galasa/pull/468/commits/97bd64e57cef8a739cb0f68ece340ff6bfe14608

PR #469, opened 2025-12-11
Make user priority settable via the REST API and CLI

eamansour:
## Why?

This PR makes the `priority` field in user records settable via the `PUT /users/{user-number}` endpoint and using the Galasa CLI tool's `galasactl users set --login-id <user-to-update> --priority <new-priority-number>`.

## Changes
- Updated the openapi2beans generator tool to use the Java `Integer` wrapper instead of the primitive `int` when generating Java beans that are used when parsing JSON payloads from requests.
  - This means that if a JSON payload omits an optional integer field, that field will be correctly parsed as `null` rather than a `0`. For example, If I wanted to update a user's role only and the user had a role "tester" and priority 100, by not including a priority value in the JSON, the server would have changed the user's priority from 100 to 0.
- Added an optional `priority` field to the JSON payload for `PUT /users/{user-number}`
  - Note: Users with the `owner` role are the only users that can update their own priority. This can be changed if needed.
- Added a `--priority` flag to the `galasactl users set` command

### DIFF
--- a/modules/cli/docs/generated/galasactl_users_set.md
+++ b/modules/cli/docs/generated/galasactl_users_set.md
@@ -15,7 +15,7 @@ galasactl users set [flags]
 ```
   -h, --help              Displays the options for the 'users set' command.
       --login-id string   A mandatory field indicating the login ID of a user.
-      --priority int      An optional field indicating the new priority of the specified user.
+      --priority int      An optional field indicating the new priority of the specified user. The higher the number, the higher the priority.
       --role string       An optional field indicating the new role of the specified user.
 ```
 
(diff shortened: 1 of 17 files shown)


modules/cli/docs/generated/galasactl_users_set.md
```diff
@@ -15,6 +15,7 @@ galasactl users set [flags]
 ```
   -h, --help              Displays the options for the 'users set' command.
       --login-id string   A mandatory field indicating the login ID of a user.
+      --priority int      An optional field indicating the new priority of the specified user.
```
jadecarino: For users who have gone straight into using this command with the `--help` text and haven't read the docs fully, might it be worth saying here "The higher the integer, the higher the priority"? I feel that it's almost self explanatory but people might misunderstand and think that a priority of `1` means you're "first in the queue" sort of thing.

jadecarino (approved):

modules/cli/docs/generated/galasactl_users_set.md
eamansour: Agreed! Added help text in https://github.com/galasa-dev/galasa/pull/469/commits/9af6dc73b26f2265991dab78d228cb783ff1d69a

PR #470, opened 2025-12-11
Issue 2245, Zosprogram manager name sanitisation

sd66dar:
### Why
This PR addresses galasa-dev/projectmanagement#2245.
The issue required enforcing two rules for @ZosProgram annotations:
The name property should always be uppercased.
The name must not exceed 8 characters; otherwise, an exception should be thrown.

### What Changed

Added logic in ZosProgramManagerImpl.generateZosProgram() to:
Convert the name attribute to uppercase using toUpperCase().
Validate the length of the name and throw a ZosProgramManagerException if it exceeds 8 characters.

### Testing

Created a static inner class Dummy with real @ZosProgram annotations to mimic production code and validate:
Names are correctly uppercased.
Names with valid lengths (≤ 8 chars) are accepted.
Names exactly 8 characters long are accepted.

Added a test for invalid names (> 8 chars) using Mockito to mock the annotation and verify that an exception is thrown.
Used Mockito instead of PowerMockito for simplicity and faster execution.
Implemented an @ After teardown method to reset the singleton instance of ZosProgramPropertiesSingleton between tests, preventing state leakage that caused earlier failures.


jadecarino (changes requested):
Hi Sheriell, thank you for making your first PR!

Changes to do with checking the 8 character limit look good, however I'd like to request a change to do with the capitalisation of the name.

The `name` value is used later on in `ZosProgramImpl` and so if we simply capitalise that value and users have their program source named in non-capital letters like `myprog.cbl`, this change will mean the manager fails to find that file now. If the `name` is not capitalised, please can we throw an exception to state that names in the annotation and file names must be capitalised? A short description of this in the documentation for this manager (found [here](https://github.com/galasa-dev/galasa/blob/main/docs/content/docs/managers/zos-managers/zos-program-manager.md)) would also be really useful. And, it would be great if you could also add a point to the v0.46.0 version's [release notes](https://github.com/galasa-dev/galasa/blob/main/docs/content/releases/posts/v0.46.0.md) in the section `Changes affecting tests running locally or on the Galasa Service` to describe the new validation checks that the z/OS Program Manager will make.

jadecarino (approved):

sd66dar:
Update:

Changed code to instead throw an exception when a name is not uppercase, and updated the tests relating to name capitalisation accordingly 
Added a description of the changes in v0.46.0 version's [release notes](https://github.com/galasa-dev/galasa/blob/main/docs/content/releases/posts/v0.46.0.md)

PR #471, opened 2025-12-11
Add priority to tags and include tag priorities in test run scheduling 

eamansour:
## Why?

This PR adds a tag service that can manipulate tags in the CPS, and allows tags to have priority values associated with them. The priority values associated with tags are then picked up if a queued test run has a tag with a priority value so that test runs with certain tags may be treated with higher priority than other test runs.

## Changes
- [x] Added a new ITagsService with CRUD operations for tags, using the CPS as the underlying storage
- [x] Tags are stored in the CPS with the format `tags.base64EncodedTagName.suffix`. The tag name is Base64 URL encoded so that we can support tags with special characters like spaces and periods.
- [x] Update the PrioritySchedulingService to collect any tags associated with queued test runs in advance before using them in calculating test run priorities

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/tags/TagTransform.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/tags/TagTransform.java
@@ -11,6 +11,8 @@ import java.util.Map;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import dev.galasa.framework.spi.tags.Tag;
+
 public class TagTransform {
 
     private final Log logger = LogFactory.getLog(getClass());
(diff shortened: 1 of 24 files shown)


modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/tags/TagTransform.java
```diff
@@ -0,0 +1,94 @@
...
+            String tagName = decodeTagName(encodedTagName);
+            if (!stringValidator.isLatin1(tagName)) {
+                logger.warn("Decoded tag name contains characters that are not in the Latin-1 character set, returning null tag");
+            } else {
+                tag = new Tag(tagName);
+    
+                String description = properties.get(TAG_DESCRIPTION_SUFFIX);
+                tag.setDescription(description);
+                String priorityString = properties.get(TAG_PRIORITY_SUFFIX);
+                int priority = 0;
```
jadecarino, line 66: Do we not need to pass `getTagPropertyKey(encodedTagName, <SUFFIX_CONSTANT>` in to `properties.get()` here instead of just the suffix constant?

jadecarino: Ah okay no worries, I see later on that the `properties` would be a grouped properties map for just one tag.

jadecarino (approved):
Looks good!

PR #473, opened 2025-12-15
Add POST /tags and PUT /tags/{tagId} endpoints to create and update tags

eamansour:
## Why?

## Changes
- [x] Added POST `/tags` to create a tag with a payload like:
    ```json
    POST /tags
    {
      "name": "tag1" // required
      "description": "my tag's description", // optional
      "priority": 100 // optional
    }
    ```
- [x] Added PUT `/tags` to create or update a tag with a payload like:
    ```json
    PUT /tags/{tagId}
    {
      "description": "a new description!", // optional
      "priority": 123 // optional
    }
    ```

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.api.tags/src/main/java/dev/galasa/framework/api/tags/internal/routes/TagsRoute.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework.api.tags/src/main/java/dev/galasa/framework/api/tags/internal/routes/TagsRoute.java
@@ -21,7 +21,6 @@ import dev.galasa.framework.api.beans.generated.TagCreateRequest;
 import dev.galasa.framework.api.common.HttpRequestContext;
 import dev.galasa.framework.api.common.InternalServletException;
 import dev.galasa.framework.api.common.MimeType;
-import dev.galasa.framework.api.common.ProtectedRoute;
 import dev.galasa.framework.api.common.QueryParameters;
 import dev.galasa.framework.api.common.ResponseBuilder;
 import dev.galasa.framework.api.common.ServletError;
@@ -36,7 +35,7 @@ import dev.galasa.framework.spi.tags.Tag;
 import dev.galasa.framework.spi.tags.TagsException;
 import dev.galasa.framework.spi.utils.StringValidator;
 
-public class TagsRoute extends ProtectedRoute {
+public class TagsRoute extends AbstractTagRoute {
 
     // Query parameters
     public static final String QUERY_PARAMETER_NAME = "name";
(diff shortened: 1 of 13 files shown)


modules/framework/galasa-parent/dev.galasa.framework.api.tags/src/test/java/dev/galasa/framework/api/tags/internal/routes/TagByNameRouteTest.java
```diff
@@ -284,4 +264,190 @@ public void testDeleteUnknownTagByNameReturnsError() throws Exception {
...
+        MockHttpServletRequest mockRequest = new MockHttpServletRequest("/" + encodedTagName, headerMap);
+        mockRequest.setMethod(HttpMethod.PUT.toString());
+        mockRequest.setPayload(requestPayload);
+
+        MockHttpServletResponse servletResponse = new MockHttpServletResponse();
+        ServletOutputStream outStream = servletResponse.getOutputStream();
+
+        // When...
+        mockServlet.init();
+        mockServlet.doGet(mockRequest, servletResponse);
```
jadecarino: Should this be a `doPut`?

jadecarino: Same for the other test cases below?

modules/framework/galasa-parent/dev.galasa.framework.api.tags/src/main/java/dev/galasa/framework/api/tags/internal/routes/TagsRoute.java
```diff
@@ -90,6 +91,49 @@ public HttpServletResponse handleGetRequest(String pathInfo, QueryParameters que
...
+        GalasaTag createdTag = transform.createTagBean(tagToSet, externalApiServerUrl);
+        String tagJson = gson.toJson(createdTag);
+        
+        logger.info("handlePostRequest() exiting");
+        return getResponseBuilder().buildResponse(request, response, MimeType.APPLICATION_JSON.toString(), tagJson, HttpServletResponse.SC_CREATED);
+    }
+
+    private Tag buildTagFromRequestPayload(TagCreateRequest requestPayload) {
+        Tag tag = new Tag(requestPayload.getname());
+        tag.setDescription(requestPayload.getdescription());
```
jadecarino: As `description` is optional in the payload should we check that `requestPayload.getdescription()` is not null here before setting it into the `tag`? Or we may get an empty/null description and might that cause issues when setting into the CPS?

jadecarino (changes requested):
Just a couple of minor comments :)

modules/framework/galasa-parent/dev.galasa.framework.api.tags/src/test/java/dev/galasa/framework/api/tags/internal/routes/TagByNameRouteTest.java
eamansour: You're right, thanks! Addressed in https://github.com/galasa-dev/galasa/pull/473/commits/b48b9bde4bab2e560060373edd9af83bb13eec46

modules/framework/galasa-parent/dev.galasa.framework.api.tags/src/main/java/dev/galasa/framework/api/tags/internal/routes/TagsRoute.java
eamansour: Good point - this wouldn't cause issues since the `description` field is null when creating the `Tag` object and the `TagsService` handles null descriptions before setting tags into the CPS, but there's no harm in adding a null check so I've added one in to make this clearer in https://github.com/galasa-dev/galasa/pull/473/commits/b48b9bde4bab2e560060373edd9af83bb13eec46

jadecarino (approved):
Looks good

PR #476, opened 2025-12-16
Add galasactl tags get command

eamansour:
## Why?

## Changes
- [x] Added `galasactl tags get --name <tag-name> [--format <output-format>]` command to get tags from a Galasa service


modules/cli/pkg/tagsformatter/tagsFormatter.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/galasa-dev/cli/pkg/galasaapi"
+)
+
+// Displays tags in the following format:
+// name                  priority description          
+// core_regression_tests 100      Core regression tests
+// my-experimental-tag   5        Experimental tests   
+// anothertag 		     10       Dummy tag            
+// Total:3
+
```
jadecarino: Should we also show an example of the yaml format here too?

jadecarino (approved):

modules/cli/pkg/tagsformatter/tagsFormatter.go
eamansour: Good point! I've moved this summary format into the `summaryFormatter.go` file and I've also added an example of the YAML format into the `yamlFormatter.go` file - see https://github.com/galasa-dev/galasa/pull/476/commits/c5f3800a6b3bfedeb8716142dbe06a07c32ddd1c

PR #480, opened 2025-12-16
Iss2389 - Able to set a user associated with a test run, which may or may not be the same as the requestor

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2389

## Changes

- [x] New action `TEST_RUN_SET_USER` available to users with the `admin` role

- [x] POST requests to the `/runs/{groupid}` endpoint accepts a `user` field in the payload
   - [x] If one is not provided, it defaults to the same as the requestor
   - [x] If the requestor has the `admin` role, they are allowed to set the run user, so this user is passed to the Framework for use in submitting the run (includes DSS updates) and storing in the test structure
   - [x] If the requestor is not an `admin`, they are not allowed to set the run user, so the `user` value is ignored and defaults to the requestor
   - [x] The `user` must be a user with the `TEST_RUN_LAUNCH` action or an error will be returned

- [x] GET requests to the `/runs/{groupid}`, `/ras/runs` and `/ras/runs/{runid}` endpoints now return a payload with the `user` field
   - [x] You can now query the `/ras/runs` endpoint with a `user` query parameter

- [x] As per the story requirements, the Mango query to CouchDB is now built up with the following rules:
   - [x] If query contains requestor only, find runs that strictly match the requestor
   - [x] If query contains user only, find runs where that user value matches either the requestor or the user
   - [x] If query contains requestor and user, find runs where the requestor strictly matches the requestor and the user matches the user

- [x] Unit tests have been updated/new test cases written for each updated servlet route
   - [x] Mock objects have also been updated, with a new constructor added to avoid having to change everywhere the Mock is used so far


modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/FrameworkRuns.java
```diff
@@ -618,6 +623,17 @@ private void setRunRequestRequestorDefault(SubmitRunRequest runRequest) throws F
         }
     }
 
+    private void setRunRequestUserDefault(SubmitRunRequest runRequest) throws FrameworkException {
+        String user = AbstractManager.nulled(runRequest.getUser());
+        if (user == null) {
+            user = AbstractManager.nulled(cps.getProperty("run", "user"));
```
eamansour: I see this comes from the `setRunRequestRequestorDefault` method above, but this seems to mean that the user can be set using a CPS property if no user was passed in - is this needed?

I'm not sure why the `setRunRequestRequestorDefault` method has that in either!

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/ras/RasSearchCriteriaUser.java
```diff
@@ -0,0 +1,51 @@
...
+    public RasSearchCriteriaUser(@NotNull String... userCriteria) {
+        this.users = userCriteria;
+
+    }
+
+    @Override
+    public boolean criteriaMatched(@NotNull TestStructure structure) {
+        
+        if (structure == null) {
+            return Boolean.FALSE;
```
eamansour: Very minor - I know this comes from other criteria classes but we can use `true`/`false` values directly instead of `Boolean.FALSE`/`Boolean.TRUE` in this method

eamansour (approved):
Looks good!

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/FrameworkRuns.java
jadecarino: Good point thanks for mentioning - upon closer inspection it looks like these methods receive a `runRequest` which when coming from a remote run through the API, will always have a `requestor` and `user` set (`user` will have defaulted to `requestor` if not provided before this point), so will never reach the block of code checking the CPS.

That block will run only for local runs which always have `null` requestor and user passed in. I can see in my RAS for local runs `requestor` and `user` are always `unknown` because I do not have those CPS props set.

For now, I'll remove the CPS property check for the `user` field and always set it to `unknown` if not in the `runRequest` as the value of the `user` field is just for remote runs where we can query the REST API - so no value for local runs.

Removed in https://github.com/galasa-dev/galasa/pull/480/commits/0947baa9087a6de08290fb37997adcf42aca0585

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/ras/RasSearchCriteriaUser.java
jadecarino: Sure, thanks, removed in https://github.com/galasa-dev/galasa/pull/480/commits/94e4a32074dc15b64c1d71f2cb65c655235afc3e

PR #482, opened 2025-12-17
Upgrade Kubernetes client-java library to v25.0.0

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2423

## Changes
- Upgraded the kubernetes client-java library that we use from v24 to v25
- Added a compatibility table to show the versions of kube that recent Galasa releases support
- Updated the set-version script in the docs module to add a new row to the compatibility table based on the most recent version of Galasa


modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/bnd.bnd
```diff
@@ -28,7 +28,7 @@ Embed-Dependency: *;scope=compile|runtime
     grpc-util-1.66.0.jar; lib:=true,\
     gson-*.jar; lib:=true,\
     guava-*.jar; lib:=true,\
-    jackson-core-2.16.1.jar; lib:=true,\
+    jackson-core-*.jar; lib:=true,\
```
jadecarino, line 31: If we are wildcarding this now and using the platform version, then can we remove the explicit version 2.16.1 from the build.gradle file of this project too please?

modules/platform/dev.galasa.platform/build.gradle
```diff
@@ -36,9 +36,9 @@ dependencies {
 
         api 'com.auth0:java-jwt:4.4.0' // used by wrapper.
         
-        api 'com.fasterxml.jackson.core:jackson-annotations:2.16.0' // used by wrapper.
-        api 'com.fasterxml.jackson.core:jackson-core:2.16.0' // used by wrapper.
-        api 'com.fasterxml.jackson.core:jackson-databind:2.16.0' // used by wrapper.
+        api 'com.fasterxml.jackson.core:jackson-annotations:2.20' // used by wrapper.
+        api 'com.fasterxml.jackson.core:jackson-core:2.20.0' // used by wrapper.
+        api 'com.fasterxml.jackson.core:jackson-databind:2.20.0' // used by wrapper.
```
jadecarino, line 42: Note - MEND scan for v0.46.0 might be ugly now we've changed the platform, will have to think of a workaround

jadecarino (changes requested):
One small change just to remove as many explicit versions as we can, and something we need to remember for next release

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/bnd.bnd
eamansour: Thanks, removed in https://github.com/galasa-dev/galasa/pull/482/commits/4429d01e25db3f29fc860a052ba5e28da6cbbb88

jadecarino (approved):

PR #485, opened 2025-12-19
Iss2389 - Set a --user with a batch of runs to submit with `galasactl runs submit`

jadecarino:
## For

https://github.com/galasa-dev/projectmanagement/issues/2389

## Changes

- [x] Support the `--user` flag on the `galasactl runs submit` command to specify a user to associate with the tests that may be different to the requestor (JWT authenticated user)
   - [x] Add new error to be returned to users if the user specified with `--user` does not have test launch permissions and inform users if they've specified a user that can't launch tests
- [x] README.md updates
- [x] CLI documentation updates
- [x] Release v0.46.0 notes updated


docs/content/docs/manage-ecosystem/runs-submit.md
```diff
@@ -103,3 +103,36 @@ galasactl runs submit `
     --override zos.default.cluster=MYPLEX `
     --log -
 ```
+
+
+### Overriding the test run 'user'
+
+When submitting runs to a Galasa ecosystem with `runs submit`, the run `requestor` will be set to the user who owns the personal access token that was used to authenticate to the Galasa ecosystem. If you wish to associate a different user with this batch of runs as the run `user`, you can use the `--user` flag. This is useful if you submit runs to a Galasa ecosystem in an automation tool or workflow and the `requestor` is a functional ID or bot account in the tool, but you wish to specify the actual user who triggered the automation, so they can query their runs later with the `runs get` command.
+
+Only users with the `admin` role can use the `--user` flag in their `runs submit` command, or the `user` will default to the authenticated requestor. The `user` specified in the command must be an existing system user, i.e., they must have access to the Galasa service and be a `requestor` of some existing test runs in the service. They must also have permission to submit test runs, i.e., have the `tester` role or more in the system.
```
eamansour: Minor - users with the `owner` role can also do this since they have the same permissions as admins

modules/cli/pkg/cmd/runsSubmit_test.go
```diff
@@ -602,6 +602,53 @@ func TestRunsSubmitTraceFlagReturnsOk(t *testing.T) {
...
+	factory := utils.NewMockFactory()
+	commandCollection, _ := setupTestCommandCollection(COMMAND_NAME_RUNS_SUBMIT, factory, t)
+
+	var args []string = []string{"runs", "submit", "--user", "mytestuser", "anothertestuser"}
+
+	// When...
+	err := commandCollection.Execute(args)
+
+	// Then...
+	assert.NotNil(t, err)
```
eamansour, line 635: We could also check what the error message contains to make sure we got the right error back

modules/cli/pkg/cmd/runsSubmit_test.go
```diff
@@ -602,6 +602,53 @@ func TestRunsSubmitTraceFlagReturnsOk(t *testing.T) {
...
+	factory := utils.NewMockFactory()
+	commandCollection, _ := setupTestCommandCollection(COMMAND_NAME_RUNS_SUBMIT, factory, t)
+
+	var args []string = []string{"runs", "submit", "--user"}
+
+	// When...
+	err := commandCollection.Execute(args)
+
+	// Then...
+	assert.NotNil(t, err)
```
eamansour, line 653: Same here - it would be great if we could check the content of the error message

modules/cli/pkg/launcher/remoteLauncher.go
```diff
@@ -90,6 +91,14 @@ func (launcher *RemoteLauncher) SubmitTestRun(
 	testRunRequest.SetOverrides(overrides)
 	testRunRequest.SetTags(tags)
 
+	// If a `--user` was not specified, the API will default
+	// it to the requestor, so no need to set it here.
+	if len(user) != 0 {
```
eamansour: Very minor - this condition could be simplified to:

```go
if user != ""
```

modules/cli/pkg/launcher/remoteLauncher.go
```diff
@@ -98,25 +107,47 @@ func (launcher *RemoteLauncher) SubmitTestRun(
...
+
+				if httpResponse != nil {
+					defer httpResponse.Body.Close()
+				}
+
+				if err != nil {
+					log.Println("SubmitTestRun - Failed to submit runs to the Galasa service")
+
+					if httpResponse.StatusCode == 403 {
+						err = galasaErrors.NewGalasaError(galasaErrors.GALASA_USER_MISSING_TEST_LAUNCH_PERMISSION, user)
```
eamansour, line 126: Not sure if this explicit 403 error code check is needed here - the API server should be returning an error message that mentions the user has insufficient privileges, so we could let the CLI display that error?

Happy to keep this as-is, just thinking about keeping the CLI code simple

eamansour (approved):
A few minor comments, but happy to approve :)

docs/content/docs/manage-ecosystem/runs-submit.md
jadecarino: Thanks Eamonn great point - addressed in https://github.com/galasa-dev/galasa/pull/485/commits/d0debbb7ab79b33ef35730a2ebb2cc85b36f613a

modules/cli/pkg/cmd/runsSubmit_test.go
jadecarino: Thanks Eamonn - addressed in https://github.com/galasa-dev/galasa/pull/485/commits/433293355174fd1a7783f07aa2d9d8b8af42a7f3

modules/cli/pkg/cmd/runsSubmit_test.go
jadecarino: Thanks Eamonn - addressed in https://github.com/galasa-dev/galasa/pull/485/commits/433293355174fd1a7783f07aa2d9d8b8af42a7f3

modules/cli/pkg/launcher/remoteLauncher.go
jadecarino: Thanks Eamonn - addressed in https://github.com/galasa-dev/galasa/pull/485/commits/886d4cf1696247c860881d99a59c0f1bb23b4bce

modules/cli/pkg/launcher/remoteLauncher.go
jadecarino: Thanks for the feedback Eamonn. I have decided to keep this as-is because if we remove these changes and then the CLI simply displays the error returned from the API, all it says is `403 Forbidden` and so I feel this informs the user in more detail what the problem is. Happy to change at a later date if users think it's too much information etc.

PR #490, opened 2025-12-23
James C CLI updates tags

James-Cocker:
# Why?
Solves [#2465](https://github.com/galasa-dev/projectmanagement/issues/2465) by adding a new command `galasactl runs update` which can be used like the following to alter tags on an existing run `galasactl runs update --name U12345 --add-tags tag1,tag2 --remove-tags tag3`.


modules/cli/docs/generated/galasactl_runs.md
```diff
@@ -33,4 +33,5 @@ Assembles, submits and monitors test runs in Galasa Ecosystem
 * [galasactl runs prepare](galasactl_runs_prepare.md)	 - prepares a list of tests
 * [galasactl runs reset](galasactl_runs_reset.md)	 - reset an active run in the ecosystem
 * [galasactl runs submit](galasactl_runs_submit.md)	 - submit a list of tests to the ecosystem
+* [galasactl runs update](galasactl_runs_update.md)	 - Update tags for a named test run.
```
eamansour: While it's true that `runs update` only updates tags at the moment, maybe this description could be made more general to something like: "Update the record of an existing test run on a Galasa service"?

It's also worth explicitly saying that this affects existing test runs on a Galasa service since I don't believe we're setting tags for local runs.

modules/cli/docs/generated/galasactl_runs_update.md
```diff
@@ -0,0 +1,35 @@
...
+Update tags for a named test run.
+
+```
+galasactl runs update [flags]
+```
+
+### Options
+
+```
+      --add-tags strings      comma-separated list of tags to add
```
eamansour: Can the --add-tags flag also be supplied like:
```
--add-tags tag1 --add-tags tag2
```

If so, it would be good to make it clear that you can either supply a comma-separated list or multiple `--add-tags` flags

modules/cli/docs/generated/galasactl_runs_update.md
```diff
@@ -0,0 +1,35 @@
...
+  -h, --help                  Displays the options for the 'runs update' command.
+      --name string           the name of the test run we want to update
+      --remove-tags strings   comma-separated list of tags to remove
```
eamansour: Can the --remove-tags flag also be supplied like:

--remove-tags tag1 --remove-tags tag2

If so, it would be good to make it clear that you can either supply a comma-separated list or multiple --remove-tags flags

modules/cli/pkg/runs/runs.go
```diff
@@ -89,13 +89,13 @@ func getRunIdFromRunName(runName string,
 	return runId, err
 }
 
-func createUpdateRunRequest(status string, result string) *galasaapi.UpdateRunRequest {
-	var UpdateRunRequest = galasaapi.NewUpdateRunRequest()
+func createUpdateRunStatusRequest(status string, result string) *galasaapi.UpdateRunRequest {
+	var UpdateRunStatusRequest = galasaapi.NewUpdateRunRequest()
```
eamansour: Very minor, can the `UpdateRunStatusRequest` variable use camel case (i.e. remove the capital U at the start :))

modules/cli/pkg/runs/runsCancel.go
```diff
@@ -175,8 +175,8 @@ func cancelByRunName(runName string, timeService spi.TimeService, commsClient ap
 
 	if err == nil {
 
-		UpdateRunRequest := createUpdateRunRequest(CANCEL_STATUS, CANCEL_RESULT)
-		err = cancelRun(runName, runId, UpdateRunRequest, commsClient)
+		UpdateRunStatusRequest := createUpdateRunStatusRequest(CANCEL_STATUS, CANCEL_RESULT)
```
eamansour: Same here, the `UpdateRunStatusRequest` variable should be in camel case

modules/cli/pkg/runs/runsReset.go
```diff
@@ -49,9 +49,9 @@ func ResetRun(
 
 		if err == nil {
 
-			UpdateRunRequest := createUpdateRunRequest(RESET_STATUS, RESET_RESULT)
+			UpdateRunStatusRequest := createUpdateRunStatusRequest(RESET_STATUS, RESET_RESULT)
```
eamansour: And here - use camel case for `UpdateRunStatusRequest`

modules/cli/pkg/runs/runsUpdate.go
```diff
@@ -0,0 +1,200 @@
...
+		addTags,
+		timeService.Now(),
+	)
+
+	var runs []galasaapi.Run
+	runs, err = GetRunsFromRestApi(runsQuery, commsClient)
+
+	if err == nil {
+		if len(runs) == 0 {
+			err = galasaErrors.NewGalasaError(galasaErrors.GALASA_ERROR_SERVER_DELETE_RUN_NOT_FOUND, runName)
```
eamansour: This error mentions `DELETE_RUN_NOT_FOUND` - should this be a different error?

modules/cli/pkg/runs/runsUpdate.go
```diff
@@ -0,0 +1,200 @@
...
+		} else {
+			err = updateRuns(runs, addTags, removeTags, commsClient)
+		}
+	} else {
+		console.WriteString(err.Error())
```
eamansour: Wondering if we need this - if the error is returned, wouldn't the CLI output the error?

If that's the case, then this might cause the same error to be printed out twice

modules/cli/pkg/runs/runsUpdate.go
```diff
@@ -0,0 +1,200 @@
...
+
+		if httpResponse != nil {
+			defer httpResponse.Body.Close()
+		}
+
+		// Non 200-299 http status codes manifest as an error.
+		if err != nil {
+			if httpResponse == nil {
+				// We never got a response, error sending it or something ?
+				err = galasaErrors.NewGalasaError(galasaErrors.GALASA_ERROR_RESET_RUN_FAILED, runName, err.Error())
```
eamansour: Looks like the wrong error message is being used here

modules/cli/pkg/runs/runsUpdate.go
```diff
@@ -0,0 +1,200 @@
...
+			} else {
+				statusCode := httpResponse.StatusCode
+				if statusCode != http.StatusAccepted && statusCode != http.StatusOK {
+					err = galasaErrors.NewGalasaError(galasaErrors.GALASA_ERROR_RESET_RUN_RESPONSE_PARSING)
```
eamansour: Looks like the wrong error message is being used here as well

modules/cli/pkg/tags/tagsSet.go
```diff
@@ -114,7 +114,7 @@ func validateTagName(tagName string) error {
 	var err error
 	log.Println("Validating the provided tag name")
 
-	err = validateStringIsLatin1AndNotBlank(tagName, galasaErrors.GALASA_ERROR_INVALID_TAG_NAME)
+	err = validateStringIsLatin1AndNotBlank(tagName, galasaErrors.GALASA_ERROR_DELETE_TAG_INVALID_NAME)
```
eamansour: This error is being used in `galasactl tags set`, but the error now says `ERROR_DELETE_TAG`. Maybe keep the error as it was previously?

modules/cli/pkg/runs/runsUpdate.go
```diff
@@ -0,0 +1,200 @@
...
+
+	addTags = removeDuplicateTags(addTags)
+	removeTags = removeDuplicateTags(removeTags)
+
+	// Check for tags that are both added and removed.
+	for _, tag := range addTags {
+		if (slices.Contains(removeTags, tag)) {
+			return galasaErrors.NewGalasaError(galasaErrors.GALASA_ERROR_UPDATE_RUN_INVALID_TAG_UPDATE, tag)
+		}
+	}
```
eamansour: Would it be worth moving this validation into its own function and then calling the validation function here?

modules/cli/docs/generated/galasactl_runs_update.md
James-Cocker: Yes it can, great point!

modules/cli/pkg/runs/runsUpdate.go
James-Cocker: Great point, looks like I didn't check over the code i grabbed from runsDelete thoroughly enough, thank you for pointing that out.

modules/cli/pkg/runs/runsUpdate.go
James-Cocker: Agreed, however I was taking inspiration from runsDelete.go where a similar thing takes place. Should i remove this print statement too?
```
                 if err != nil {
			console.WriteString(err.Error())
		}
	}
	log.Printf("RunsDelete exiting. err is %v\n", err)
	return err
```

eamansour: Since this PR is focused on tags, I'd say leave the `runs delete` code as-is and we can raise a separate PR for that if needed.

modules/cli/pkg/tags/tagsSet.go
James-Cocker: I see, It was under the tags delete section in errorMessage.go but i should have cross referenced it before changing its name. I still think it's important to be specific to the method it's being used in, and not just 'INVALID_NAME_TAG'. This way, it's more obvious that the error is focused on the field "name", where it could be mistaken for the name of the tag itself.

I've changed the error name to `GALASA_ERROR_SET_TAG_INVALID_NAME` with this in mind, is that okay or should I still reverse it?

modules/cli/pkg/errors/errorMessage.go
```diff
@@ -480,6 +479,7 @@ var (
 	GALASA_ERROR_SET_TAG_SERVER_REPORTED_ERROR       = NewMessageType("GAL1262E: Failed to set a tag named '%s'. Unexpected http status code %v received from the server. Error details from the server are: '%s'", 1262, STACK_TRACE_NOT_WANTED)
 	GALASA_ERROR_SET_TAG_EXPLANATION_NOT_JSON        = NewMessageType("GAL1263E: Failed to set a tag named '%s'. Unexpected http status code %v received from the server. Error details from the server are not in the json format.", 1263, STACK_TRACE_NOT_WANTED)
 	GALASA_ERROR_SET_TAG_REQUEST_FAILED              = NewMessageType("GAL1264E: Failed to set a tag named '%s'. Sending the put request to the Galasa service failed. Cause is %v", 1264, STACK_TRACE_NOT_WANTED)
+	GALASA_ERROR_SET_TAG_INVALID_NAME             	 = NewMessageType("GAL1256E: Invalid tag name provided. The name provided with the --name flag cannot be empty and must only contain characters in the Latin-1 character set.", 1256, STACK_TRACE_NOT_WANTED)
```
eamansour: Just had a look at the code that uses this and it's also used by the `tags get` and `tags delete` commands, so maybe reverting this back to `GALASA_ERROR_INVALID_TAG_NAME` would be most accurate?

modules/cli/pkg/runs/runsUpdate.go
```diff
@@ -179,6 +173,24 @@ func updateRun(
...
+	var err error
+
+	if runName == "" {
+		return galasaErrors.NewGalasaError(galasaErrors.GALASA_ERROR_MISSING_NAME_FLAG, "--name")
+	}
+
+	if len(addTags) == 0 && len(removeTags) == 0 {
+		err = galasaErrors.NewGalasaError(galasaErrors.GALASA_ERROR_UPDATE_RUN_MISSING_FIELD, "--add-tags or --remove-tags")
+		return err
+	}
```
eamansour, line 190: Do we also want to validate that the supplied tags aren't made up of empty strings/contain newline characters?

Perhaps we should also trim the tags to remove any starting/trailing spaces?

eamansour (approved):

PR #492, opened 2025-12-24
Add the ability to run individual test methods

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2497

## Changes
- [x] Added a new `--methods` flag to the Galasa boot jar which accepts Galasa test method names that a user wishes to run
- [x] The TestClassWrapper now receives a list of test methods that should be run (which includes befores + afters). As it loops through test methods in a class, any test methods that are not in the list of methods to run will be marked as ignored (including their befores + after methods).
- [x] Added a new `--methods` flag to the `galasactl runs submit local` command which passes a set of flags to the boot jar launch command
- [x] Updated release notes and `galasactl runs submit local` command docs

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/init/TestRunInitStrategy.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/init/TestRunInitStrategy.java
@@ -5,6 +5,8 @@
  */
 package dev.galasa.framework.internal.init;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Properties;
 import java.util.Set;
 import java.util.UUID;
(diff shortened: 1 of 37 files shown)


modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/init/TestRunInitStrategy.java
```diff
@@ -56,17 +59,36 @@ private String locateRunName(IFramework framework ,IConfigurationPropertyStoreSe
...
             }
             logger.info("Submitting test "+testName);
-            runName = submitRun(framework, testName, testLanguage);
+            runName = submitRun(framework, testName, testLanguage, requestedTestMethods);
         }
         logger.info("Run name is "+runName);
         return runName;
     }
 
+    private List<String> getRequestedTestMethods(IConfigurationPropertyStoreService cps) throws ConfigurationPropertyStoreException {
```
jadecarino, line 75: Should we add into the docs that you can control the methods using CPS as well as the `--methods` flag or, do we not really want to advertise this if it's just here to be consistent with the rest of this class?

jadecarino: As I doubt anyone would set into CPS `run.testmethods` (or `run.testbundleclass` for that matter) before kicking off a run

jadecarino: Oh right I see - this would be added by the Launcher. Ignore me!

jadecarino (approved):
Looks good!

PR #497, opened 2026-01-09
Improve CouchDB query performance for single-value filters

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2505

## Changes
- [x] Removed the deprecated page-based pagination from the `/ras/runs` endpoint to take advantage of cursor-based pagination performance enhancements by default

- [x] Deprecated the `includeCursor` query parameter as it is no longer needed

- [x] Updated the CouchDB query builder code to use equality comparisons for single-value fields instead of using the `$in` operator for all comparisons.
  - Multi-valued fields will still work as they did before

For example, when querying `/ras/runs?group=my-group&result=failed`, the previous CouchDB query would look like this:
```json
{
  "selector": {
    "$and": [
      {
        "group": {
          "$in": [
            "my-group"
          ]
        }
      },
      {
        "result": {
          "$in": [
            "Failed"
          ]
        }
      }
    ]
  }
}
```

Now, it looks like this:
```json
{
  "selector": {
    "$and": [
      {
        "group": "my-group"
      },
      {
        "result": "Failed"
      }
    ]
  }
}
```

- [x] Added new indexes with multiple fields to improve performance for queries that use the following fields:
  - [x] requestor and from/to
  - [x] requestor, user, and from/to
  - [x] result and from/to
  - [x] status and from/to
  - [x] group and result
  - [x] group and status

- [x] Stopped emitting entire documents in the group view to save disk space (the recommendation is to use `include_docs=true` in view queries to get documents back, which is what we already do)

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
@@ -68,6 +68,9 @@ public class RunQueryRoute extends RunsRoute {
 			"to", "endTime",
 			"testclass", "testName");
 
+	// No longer used, but keep it for now to avoid breaking existing clients
+	public static final String QUERY_PARAMETER_INCLUDECURSOR = "includecursor";
+
 	public static final String QUERY_PARAMETER_SORT = "sort";
 	public static final String QUERY_PARAMETER_RESULT = "result";
 	public static final String QUERY_PARAMETER_STATUS = "status";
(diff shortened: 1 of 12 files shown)


modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RunQueryRoute.java
```diff
@@ -156,82 +146,24 @@ private String retrieveResults(RasQueryParameters queryParams, boolean isMethodD
...
-						String runName = queryParams.getRunName();
-						if (runName != null) {
-							runsPage = new RasRunResultPage(getRunsByRunName(runName));
-						} else {
-							runsPage = getRunsPage(pageCursor, pageSize, formatSortField(sortValue), criteria);
-						}
-					} else {
-						runs = getRuns(criteria, isMethodDetailsExcluded);
-					}
+				UserQueryContext userQueryContext = matchRequestorAndUser(
```
jadecarino, line 154: I like this change, thanks!

jadecarino (approved):
Looks good!

PR #500, opened 2026-01-12
Added small docs changes

James-Cocker:
# Why?
Picks up on the description in the `openapi.yaml` file which should have been changed to reflect it's new capabilities of adding tags. Includes the edit of the operation ID.

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
+++ b/modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
@@ -1796,7 +1796,7 @@ paths:
 
         Requests to this endpoint require a valid bearer token in JWT format to be provided
         in the 'Authorization' header (e.g. 'Authorization: Bearer <bearer-token>').
-      operationId: putRasRunStatusById
+      operationId: putRasRunTagsOrStatusById
       tags:
       - Result Archive Store API
       parameters:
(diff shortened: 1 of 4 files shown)


modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -1780,21 +1780,23 @@ paths:
           $ref: '#/components/responses/InternalServerError'
 
   #--------------------------------------
-  # Update the status of an active test
-  # run to reset or cancel it.
+  # Update the tags or status of an active
+  # test run to reset or cancel it.
```
jadecarino: I feel like this is a little confusing and could seem as though updating the tags can reset/cancel a run. Please can we update to be a bit clearer? Something like `Update the tags associated with a test run, or update the status to reset or cancel it.`

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -1780,21 +1780,23 @@ paths:
...
   #--------------------------------------
     put:
-      summary: Update the status of a test run
+      summary: Update the tags, status or result of a test run
       description: |
-        Updates the status of a test run in order to either reset or cancel it.
+        Updates the tags, status or result of a test run in order to either change the tags or reset/ cancel it.
```
jadecarino: Similar thoughts as to above, can we please change this first sentence to have two parts if you like to make it clear that its two separate things?

jadecarino (changes requested):

jadecarino (approved):

PR #505, opened 2026-01-19
Iss2312 - Pass max gRPC message size from Helm chart into ETCD client creation

jadecarino:
## Why?

## Changes

- Environment variable `MAX_GRPC_MESSAGE_SIZE` if passed in from Helm chart is used to create the ETCD Client for DSS with `maxInboundMessageSize()`
- If it is not provided in the environment, it is set to the default value 4Mib
- Validation of the environment variable in case of letters/whitespace/negative numbers being passed in
- Unit tests to ensure ETCD registration and store can be set up okay with this new parameter, and to test the defaulting logic for if not provided/invalid value is provided

### DIFF
--- a/modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStoreRegistration.java
+++ b/modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStoreRegistration.java
@@ -30,8 +30,8 @@ import dev.galasa.framework.spi.SystemEnvironment;
 @Component(service = { IDynamicStatusStoreRegistration.class })
 public class Etcd3DynamicStatusStoreRegistration implements IDynamicStatusStoreRegistration {
 
-    private static final String MAX_GRPC_MESSAGE_SIZE_ENV_VAR = "MAX_GRPC_MESSAGE_SIZE";
-    private static final int DEFAULT_MAX_GRPC_MESSAGE_SIZE = 4194304;
+    public static final String MAX_GRPC_MESSAGE_SIZE_ENV_VAR = "MAX_GRPC_MESSAGE_SIZE";
+    public static final int DEFAULT_MAX_GRPC_MESSAGE_SIZE = 4194304;
 
     private final Log logger = LogFactory.getLog(Etcd3DynamicStatusStoreRegistration.class);
 
--- a/modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStoreRegistration.java
+++ b/modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStoreRegistration.java
@@ -66,7 +66,7 @@ public class Etcd3DynamicStatusStoreRegistration implements IDynamicStatusStoreR
         if (isEtcdUri(dss)) {
             try {
                 URI uri = new URI(dss.toString().substring(5));
-                setMaxgRPCMessageSizeFromEnvironmentOrDefault(MAX_GRPC_MESSAGE_SIZE_ENV_VAR, DEFAULT_MAX_GRPC_MESSAGE_SIZE);
+                setMaxgRPCMessageSizeFromEnvironmentOrDefault();
                 frameworkInitialisation.registerDynamicStatusStore(new Etcd3DynamicStatusStore(uri, this.maxgRPCMessageSize));
             } catch (URISyntaxException e) {
                 throw new DynamicStatusStoreException("Could not create URI", e);
@@ -84,8 +84,8 @@ public class Etcd3DynamicStatusStoreRegistration implements IDynamicStatusStoreR
         return "etcd".equals(uri.getScheme());
     }
 
-    private void setMaxgRPCMessageSizeFromEnvironmentOrDefault(String envVar, int defaultValue) {
-        String value = this.env.getenv(envVar);
+    private void setMaxgRPCMessageSizeFromEnvironmentOrDefault() {
+        String value = this.env.getenv(MAX_GRPC_MESSAGE_SIZE_ENV_VAR);
         if (value != null && !value.isBlank()) {
             try {
                 int parsed = Integer.parseInt(value.trim());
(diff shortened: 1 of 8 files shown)


modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStoreRegistration.java
```diff
@@ -44,7 +66,8 @@ public void initialise(@NotNull IFrameworkInitialisation frameworkInitialisation
         if (isEtcdUri(dss)) {
             try {
                 URI uri = new URI(dss.toString().substring(5));
-                frameworkInitialisation.registerDynamicStatusStore(new Etcd3DynamicStatusStore(uri));
+                setMaxgRPCMessageSizeFromEnvironmentOrDefault(MAX_GRPC_MESSAGE_SIZE_ENV_VAR, DEFAULT_MAX_GRPC_MESSAGE_SIZE);
```
Mark-J-Lawrence: Does `setMaxgRPCMessageSizeFromEnvironmentOrDefault()` need to accept any arguments?

The name of the method suggests its only to work with MaxgRPCMessageSize, what else could be passed in if it were not these two constants?

Those arguments being provided are already provided at a class level, so they could just be directly used inside of `setMaxgRPCMessageSizeFromEnvironmentOrDefault`.

Or `setMaxgRPCMessageSizeFromEnvironmentOrDefault` could be named to something more generic, so any var name & default value could be provided, which has nothing to do with `MaxgRPCMessageSize`

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStoreRegistration.java
```diff
@@ -60,4 +83,26 @@ public void initialise(@NotNull IFrameworkInitialisation frameworkInitialisation
...
         return "etcd".equals(uri.getScheme());
     }
+
+    private void setMaxgRPCMessageSizeFromEnvironmentOrDefault(String envVar, int defaultValue) {
+        String value = this.env.getenv(envVar);
+        if (value != null && !value.isBlank()) {
+            try {
+                int parsed = Integer.parseInt(value.trim());
+                if (parsed >= 0) {
+                    this.maxgRPCMessageSize = parsed;
```
Mark-J-Lawrence, line 93: linking back to my last comment - this method is really locked to just being for `MaxgRPCMessageSize`, therefore, it doesn't really need the method parameters

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3Store.java
```diff
@@ -39,6 +39,8 @@
  */
 public abstract class Etcd3Store {
 
+    private static final int DEFAULT_MAX_GRPC_MESSAGE_SIZE = 4194304;
```
Mark-J-Lawrence: This constant was declared in `Etcd3DynamicStatusStoreRegistration` as well, this might introduce bugs should this value ever need updating & someone forgets to update the other.

Could one of them be public and the other java class points to the other - or could it live in a separate file and be imported in?

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/test/java/dev/galasa/etcd/internal/Etcd3DynamicStatusStoreRegistrationTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+import org.junit.Test;
+
+import dev.galasa.cps.etcd.internal.Etcd3DynamicStatusStoreRegistration;
+import dev.galasa.extensions.common.mocks.MockEnvironment;
+import dev.galasa.extensions.common.mocks.MockFrameworkInitialisation;
+
+public class Etcd3DynamicStatusStoreRegistrationTest {
+
+    private static final String MAX_GRPC_MESSAGE_SIZE_ENV_VAR = "MAX_GRPC_MESSAGE_SIZE";
```
Mark-J-Lawrence: does the test need to declare these separately as well? 

Both these constants should be declared in one place, and they're imported in

Mark-J-Lawrence (changes requested):

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStoreRegistration.java
jadecarino: Yes great point thanks Mark. Originally I wrote this method to be generic in case there were any future environment variables that we should get as an int in this class, and it started out as just get me the environment variable as an int if it was set. But then the internals of the method became quite specific to maxgRPCMessageSize to ensure if a negative integer was passed in this was handled correctly etc so I changed the method name. The arguments can be removed now. I've fixed this in https://github.com/galasa-dev/galasa/pull/505/commits/fcc11cea3b2c91eb14020bed0d649d3f39664472

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStoreRegistration.java
jadecarino: Thanks - fixed in https://github.com/galasa-dev/galasa/pull/505/commits/fcc11cea3b2c91eb14020bed0d649d3f39664472

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3Store.java
jadecarino: Good suggestion thanks I've kept these in just `Etcd3DynamicStatusStoreRegistration ` and imported them into other classes that use them. Fixed in https://github.com/galasa-dev/galasa/pull/505/commits/1a1b6172e81fa80c787727f5f0eaf06f8d55cca7

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/test/java/dev/galasa/etcd/internal/Etcd3DynamicStatusStoreRegistrationTest.java
jadecarino: Thanks Mark, fixed in https://github.com/galasa-dev/galasa/pull/505/commits/1a1b6172e81fa80c787727f5f0eaf06f8d55cca7

Mark-J-Lawrence (approved):
Thanks for resolving the comments @jadecarino - looks great!

PR #506, opened 2026-01-20
Iss2516 - Fix getAllRuns exiting with exception when gRPC max message size hit by processing ETCD key/values with pagination

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2516

## Changes

- Private method `getPropertiesFromETCDWithPrefix` which is called by `FrameworkRuns.getAllRuns()` now grabs key/values from ETCD in pages of 10,000 to attempt to avoid the response from ETCD exceeding the maximum gRPC message size.
- Tested on Minikube substituting pageSize = 1 for testing purposes and compared response without pagination to response with to confirm the same number of key/value properties were retrieved.

### DIFF
--- a/modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStore.java
+++ b/modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStore.java
@@ -9,6 +9,7 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 
 import java.net.URI;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStore.java
```diff
@@ -373,23 +374,51 @@ public Collection<String> getPrefixKeysOnly(@NotNull String keyPrefix) throws Dy
...
-        List<KeyValue> kvs = new ArrayList<>();
-        try {
-            GetResponse response = getFuture.get();
-            kvs = response.getKvs();
-        } catch (InterruptedException | ExecutionException e) {
-            Thread.currentThread().interrupt();
-            throw new DynamicStatusStoreException("Could not retrieve key.", e);
+            allKvs.addAll(kvs);
+
+            if (kvs.size() < pageSize) {
```
Mark-J-Lawrence, line 407: I guess there could be an odd chance that `kvs` could come back with 10000 records, and the 10,000 record is the last one....

I think this code would go round and make one more call and return nothing, or perhaps the startKey might not be able to be figured out and an exception is thrown?

I'll approve - but this should be considered & a fix go in if required

Mark-J-Lawrence (approved):

modules/extensions/galasa-extensions-parent/dev.galasa.cps.etcd/src/main/java/dev/galasa/cps/etcd/internal/Etcd3DynamicStatusStore.java
jadecarino: Thanks for your review Mark. I don't think there should be any issues setting the `startKey` as its simply appending `0x00` onto the last record's bytes, and then the next loop would return an empty response. I tested this on my Minikube with a pageSize of 1 so in this scenario the loop would keep going 1 record at a time until there was an empty response, and this was fine with no exceptions.

PR #507, opened 2026-01-21
Iss2519 - Update links to Javadoc in Galasa Mkdocs site to local paths

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2519

## Changes

- Updated any links within the Galasa Mkdocs site from `https://javadoc.galasa.dev/*` to `./docs/reference/javadoc/*`
   - The live Mkdocs site will render the same
   - This change will also mean that the links to the Javadoc provided in the static documentation site in the Isolated/MVP zips resolve to local files rather than navigating to the served up Javadoc site which would require internet


jadecarino:
Merging as admin as all Galasa devs on annual leave.

PR #513, opened 2026-01-29
Iss2523 - Remove httpclient and httpmine from MVP zip

jadecarino:
## Why?

For https://github.com/galasa-dev/projectmanagement/issues/2523

## Changes

- [x] Update the release.yaml to have mvp: false for httpclient and httpmime to remove these two from the MVP zip.
- [x] Update HTTP Manager (as its provided in the MVP) to depend only on dev.galasa.wrapping.httpclient-osgi and not httpclient and httpmime as they are provided in the wrapper.


jadecarino:
@ideeley FYI

PR #521, opened 2026-02-10
James C iss2455 Check client-java Version

James-Cocker:
# Why?

Closes `Part 1` of https://github.com/orgs/galasa-dev/projects/3/views/1?pane=issue&itemId=135219458&issue=galasa-dev%7Cprojectmanagement%7C2455 by adding a new workflow that runs at 9am UTC which will raise a new issue if Galasa's current `client-java` package version is out of date. Note: if a previous issue is still open and an even newer version is found then the issue will be updated, adding a comment when it does.


.github/workflows/check-client-java-version.yaml
```diff
@@ -0,0 +1,133 @@
...
+              issue.title.includes('io.kubernetes:client-java')
+            );
+
+            const issueBody = `A new version of the Kubernetes client-java is available.\n\n` +
+                  `**Current version:** ${current}\n` +
+                  `**Latest version:** ${latest}\n\n` +
+                  `**Maven Central:** https://central.sonatype.com/artifact/io.kubernetes/client-java/${latest}\n\n` +
+                  `## Tasks\n` +
+                  `- [ ] Ensure [client-java ${latest}](https://central.sonatype.com/artifact/io.kubernetes/client-java/${latest}) exists on Maven Central\n` +
+                  `- [ ] Review and update the dependency in \`modules/platform/dev.galasa.platform/build.gradle\`\n` +
```
eamansour, line 89: This is great - the build.gradle does have a few client-java dependencies that are grouped together like:
```gradle
        api 'io.kubernetes:client-java:25.0.0'
        api 'io.kubernetes:client-java-api:25.0.0'
        api 'io.kubernetes:client-java-proto:25.0.0'
```

So perhaps we should update the second task to say that we should review and update the io.kubernetes dependencies in the build.gradle file?

eamansour (approved):

.github/workflows/check-client-java-version.yaml
James-Cocker: I'll add it now 😁

PR #529, opened 2026-02-17
Only load requested artifacts once when downloading runs instead of loading all artifacts for every file download

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2541

Note: this PR builds off of changes in https://github.com/galasa-dev/galasa/pull/527 so this branch will need to be rebased.

## Changes
- [x] Updated the RAS file download endpoint to only load an artifact from CouchDB once for each request rather than loading all artifacts when requesting to download a single file


jadecarino (approved):
Looks good!

PR #537, opened 2026-02-19
Add POST /streams and PUT /streams/{streamName} to allow users to create and update test streams via the REST API

eamansour:
## Why?
For https://github.com/galasa-dev/projectmanagement/issues/2447

## Changes
- [x] Added `POST /streams` to create test streams (will error if a test stream with the given name already exists)
- [x] Added `PUT /streams/{streamName}` to update a test stream if it already exists, or create a test stream otherwise


jadecarino (approved):
All looks great!

PR #540, opened 2026-02-20
Add null handling to RunFinishedRuns, perform atomic DSS updates when moving runs to the 'finished' state

eamansour:
## Why?
Refer to https://github.com/galasa-dev/projectmanagement/issues/2516 

## Changes
- [x] Added null handling to the RunFinishedRuns class to avoid NullPointerExceptions from being thrown
- [x] When moving runs to the 'finished' state, DSS updates should now be performed atomically rather than via multiple DSS set operations to avoid situations where a run's status is marked as 'finished' but doesn't have a 'finished' timestamp property set
- [x] Added unit tests to the RunFinishedRuns class

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
@@ -11,6 +11,7 @@ import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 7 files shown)


modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/main/java/dev/galasa/framework/k8s/controller/TestPodScheduler.java
```diff
@@ -203,10 +204,13 @@ protected void startPod(IRun run) {
...
-                    this.dss.put("run." + run.getName() + "."+DssPropertyKeyRunNameSuffix.RESULT, "EnvFail" );
-                    this.dss.put("run." + run.getName() + "."+DssPropertyKeyRunNameSuffix.STATUS, "finished");
+                    Map<String, String> propertiesToSet = new HashMap<>();
+                    propertiesToSet.put("run." + run.getName() + "."+DssPropertyKeyRunNameSuffix.RESULT, "EnvFail" );
+                    propertiesToSet.put("run." + run.getName() + "."+DssPropertyKeyRunNameSuffix.STATUS, "finished");
                     Instant finishedTimeStamp = timeService.now();
-                    this.dss.put("run." + run.getName() + "."+DssPropertyKeyRunNameSuffix.FINISHED_DATETIME , finishedTimeStamp.toString());
+                    propertiesToSet.put("run." + run.getName() + "."+DssPropertyKeyRunNameSuffix.FINISHED_DATETIME , finishedTimeStamp.toString());
+
+                    this.dss.put(propertiesToSet);
```
James-Cocker, line 213: Do you think you could explain why we're using put() and not putAll() please

eamansour: The DSS service interface doesn't have a `putAll()` method. It has `put()` which is overloaded a few times, and one of them accepts a map of key-value strings to put all properties in the map into the DSS at once.

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/RunFinishedRuns.java
```diff
@@ -74,14 +75,23 @@ public void run() {
...
                 Instant now = Instant.now();
-                if (expires.compareTo(now) <= 0) {
+                if (expires == null || expires.compareTo(now) <= 0) {
                     logger.info("The run has expired so we will attempt to delete it");
-                    String sFinished = dtf.format(LocalDateTime.ofInstant(finished, ZoneId.systemDefault()));
-                    /// TODO put time management into the framework
-                    logger.info("Deleting run " + runName + ", finished at " + sFinished);
+                    if (finished != null) {
+                        String sFinished = dtf.format(LocalDateTime.ofInstant(finished, ZoneId.systemDefault()));
+                        /// TODO put time management into the framework
```
James-Cocker: Curious if the TODO is still needed

James-Cocker (approved):

modules/framework/galasa-parent/dev.galasa.framework.resource.management/src/main/java/dev/galasa/framework/resource/management/internal/RunFinishedRuns.java
eamansour: Removed in https://github.com/galasa-dev/galasa/pull/540/commits/79d872e61557a58cf330f560520b70f18833a21a

PR #543, opened 2026-02-25
Add retry loop when loading OBRs for test runs 

eamansour:
## Why?
Refer to https://github.com/galasa-dev/projectmanagement/issues/2550. 

## Changes
- [x] When the test runner loads OBRs, it can now retry the loading operation up to 10 times with a 3 second delay between retries instead of shutting down immediately
- [x] Unit tests

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/runner/FelixRepoAdminOBRAdder.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/runner/FelixRepoAdminOBRAdder.java
@@ -81,9 +81,7 @@ public class FelixRepoAdminOBRAdder {
                     logger.info("Retrying OBR load for " + obr + " (attempt " + attempt + " of " + maxRetries + ")");
                 }
                 repoAdmin.addRepository(obr);
-                if (attempt > 1) {
-                    logger.info("Successfully loaded OBR " + obr + " on retry attempt " + attempt);
-                }
+                logger.info("Successfully loaded OBR " + obr);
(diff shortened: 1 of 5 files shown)


modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/runner/FelixRepoAdminOBRAdder.java
```diff
@@ -11,21 +11,32 @@
 import dev.galasa.framework.TestRunException;
 import dev.galasa.framework.spi.AbstractManager;
 import dev.galasa.framework.spi.IConfigurationPropertyStoreService;
+import dev.galasa.framework.spi.utils.ITimeService;
 
 public class FelixRepoAdminOBRAdder {
 
+    public static final int MAX_ADD_OBR_OPERATION_RETRIES = 10;
+    public static final long DELAY_BETWEEN_ADD_OBR_OPERATION_RETRIES_MILLISECS = 3000;
```
James-Cocker, line 19: Do we have a constants file for these variables like in the webui?

James-Cocker (approved):

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/internal/runner/FelixRepoAdminOBRAdder.java
eamansour: No, these constants are specific to this class and aren't intended to be used anywhere else so moving them to a dedicated constants file could be confusing!

PR #545, opened 2026-02-25
CLI: Show the web-ui-url field only for remote runs and populate web-ui-url and tags in summary output

eamansour:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2553. 

## Changes
- [x] The `web-ui-url` field no longer appears for local test runs
- [x] The `web-ui-url` field now gets populated when remote test runs finish
- [x] The `tags` field also gets populated when both local and remote runs finish
- [x] Unit tests

### DIFF
--- a/modules/cli/pkg/runsformatter/summaryFormatter.go
+++ b/modules/cli/pkg/runsformatter/summaryFormatter.go
@@ -45,7 +45,20 @@ func (*SummaryFormatter) FormatRuns(testResultsData []FormattableTest) (string, 
 	if totalResults > 0 {
 		var table [][]string
 
-		var headers = []string{HEADER_SUBMITTED_TIME, HEADER_RUNNAME, HEADER_REQUESTOR, HEADER_STATUS, HEADER_RESULT, HEADER_TEST_NAME, HEADER_GROUP, HEADER_TAGS, HEADER_WEBUI_URL}
+		includeWebUiUrl := false
+		for _, run := range testResultsData {
+			if !run.IsLocal {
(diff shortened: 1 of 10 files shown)


modules/cli/pkg/runsformatter/summaryFormatter.go
```diff
@@ -45,7 +45,20 @@ func (*SummaryFormatter) FormatRuns(testResultsData []FormattableTest) (string,
 	if totalResults > 0 {
 		var table [][]string
 
-		var headers = []string{HEADER_SUBMITTED_TIME, HEADER_RUNNAME, HEADER_REQUESTOR, HEADER_STATUS, HEADER_RESULT, HEADER_TEST_NAME, HEADER_GROUP, HEADER_TAGS, HEADER_WEBUI_URL}
+		includeWebUiUrl := false
+		for _, run := range testResultsData {
```
James-Cocker, line 49: Looks like this says if any items in testResultsData aren't local, then they should all include a webuiurl, however i thought you couldn't mix local and non-local tests together. Is that right?

eamansour: That's right - we don't have local and non-local tests in summary outputs right now. Happy to remove this but it currently works as intended, so maybe it's worth keeping in?

James-Cocker: No worries - just checking it was intended 😁

James-Cocker (approved):

PR #546, opened 2026-02-26
James c iss2536 ras health endpoint

James-Cocker:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2536. 

Added new api route `/ras/health` to replicate the result of CouchDB's `_GO/` request.

## Changes
- [x] New API route `/ras/health`

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RasHealthRoute.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RasHealthRoute.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright contributors to the Galasa project
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package dev.galasa.framework.api.ras.internal.routes;
+
+import static dev.galasa.framework.api.common.ServletErrorMessage.*;
+
+import java.io.IOException;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import com.google.gson.JsonObject;
+
+import dev.galasa.framework.api.common.BaseRoute;
+import dev.galasa.framework.api.common.HttpRequestContext;
+import dev.galasa.framework.api.common.InternalServletException;
+import dev.galasa.framework.api.common.QueryParameters;
+import dev.galasa.framework.api.common.ResponseBuilder;
+import dev.galasa.framework.api.common.ServletError;
+import dev.galasa.framework.spi.FrameworkException;
+import dev.galasa.framework.spi.IFramework;
+import dev.galasa.framework.spi.IResultArchiveStoreDirectoryService;
+import dev.galasa.framework.spi.rbac.BuiltInAction;
+
+/**
+ * Route to check the health status of the Result Archive Store (RAS).
+ * This endpoint is publicly accessible (no authentication required).
+ */
+public class RasHealthRoute extends BaseRoute {
+
+    protected static final String path = "\\/health\\/?";
+
+    private IFramework framework;
+
+    public RasHealthRoute(ResponseBuilder responseBuilder, IFramework framework) {
+        /*
+         * Regex to match endpoints:
+         * -> /ras/health
+         * -> /ras/health/
+         */
+        super(responseBuilder, path);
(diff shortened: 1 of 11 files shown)


modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RasHealthRoute.java
```diff
@@ -0,0 +1,87 @@
...
+        try {
+            for (IResultArchiveStoreDirectoryService directoryService : framework.getResultArchiveStore()
+                    .getDirectoryServices()) {
+                if (directoryService.isHealthy()) {
+                    return true;
+                }
+            }
+            return false;
+        } catch (Exception e) {
+            return false;
```
eamansour: Very minor - we could have a single `return false;` statement outside the try-catch perhaps?

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RasHealthRoute.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    /**
+     * Checks if the RAS is healthy by querying all directory services.
+     * 
+     * @return true if at least one RAS directory service is healthy, false
+     *         otherwise
+     */
+    private boolean checkRasHealth() {
```
eamansour: Maybe rename this to `isRasHealthy()` so it's more obvious that it returns a boolean?

modules/framework/galasa-parent/dev.galasa.framework/src/testFixtures/java/dev/galasa/framework/mocks/MockResultArchiveStoreDirectoryService.java
```diff
@@ -21,6 +21,11 @@ public class MockResultArchiveStoreDirectoryService implements IResultArchiveSto
 
     private List<IRunResult> runResults;
     private String nextCursor;
+		private boolean isHealthy;
+
+		public void setHealthy(boolean isHealthy) {
+			this.isHealthy = isHealthy;
+		}
```
eamansour: Minor - can the `setHealthy` method be moved below the constructor (perhaps as close to the `isHealthy` method as possible) to keep this file consistent with the rest of the codebase?

eamansour (approved):

modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RasHealthRoute.java
James-Cocker: Hey Eamonn, like this?

```private boolean checkRasHealth() {
        try {
            for (IResultArchiveStoreDirectoryService directoryService : framework.getResultArchiveStore()
                    .getDirectoryServices()) {
                if (directoryService.isHealthy()) {
                    return true;
                }
            }
        } catch (Exception e) { }
        return false;
    }

Comment (eamansour) [reply modules/framework/galasa-parent/dev.galasa.framework.api.ras/src/main/java/dev/galasa/framework/api/ras/internal/routes/RasHealthRoute.java]: Does anything inside the try block actually throw an exception? If not, then we could remove the try-catch from the method, making it:

```java
        boolean isHealthy = false;
        for (IResultArchiveStoreDirectoryService directoryService : framework.getResultArchiveStore()
                .getDirectoryServices()) {
            if (directoryService.isHealthy()) {
                isHealthy = true;
                break;
            }
        }
        return isHealthy;
```

James-Cocker: I believe there is a `ResultArchiveStoreException` thrown, do you think it's best for that error to propagate up the Get request or have any errors simply return false for isHealthy?

eamansour: Thanks for confirming - in that case, what you sent in https://github.com/galasa-dev/galasa/pull/546#discussion_r2860086473 would be fine!

eamansour (approved):

PR #551, opened 2026-03-19
Iss2509 - Store KeyStore credentials in the CREDS, provide to HTTP/Docker Manager

jadecarino:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2509

This pull request adds the ability to store a KeyStore credential in the local and ETCD CREDS stores. The purpose of this is to be able to provide that credential to the HTTP Manager so it can use it to set up client authentication with the enclosed certificates. The HTTP Manager can then pass this to the Docker Manager, so it can contact protected Docker engines via HTTPS.

### Framework/Credentials
- [x] Adds a new credential type able to be stored in the credentials stores and used by Galasa tests, CredentialsKeyStore, which represents a Java KeyStore.
- [x] The CredentialsKeyStore has three parts: the KeyStore data which should be formatted as a prefix of "base64:" followed by the bytes of a Java KeyStore base64-encoded, the KeyStore password, and the KeyStore type (PKCS12 or JKS).
- [x] Unit tests for CredentialsKeyStore class.

### Managers
- [x] New CPS property for the Docker Manager to provide a credentials ID for a particular Docker engine. If provided, the Docker Manager will attempt to retrieve the CredentialsKeyStore from the credentials.properties, and configure an HTTPS connection with the included certificates.
- [x] Changes to the HTTP Manager to set up client authentication using this KeyStore's certificates.
- [x] Changes to DockerExecImpl as this uses a raw HttpConnection instead of the manager's HTTPClient, and so couldn't access the KeyStore certificates, so this change checks if an SSL Context is available, and if so uses it in the HttpConnection.
- [x] DockerManagerIVT has been successfully ran against a protected Docker engine to test these changes.

### REST API
- [x] KeyStore credentials can be stored and retrieved in the ETCD CREDS store by REST API calls.
- [x] Unit tests for servlet calls to create/update KeyStore credentials.
- [x] REST API changes tested in Minikube system.

### Documentation
- [x] User facing documentation.
- [x] Release notes.


docs/content/docs/managers/cloud-managers/docker-tls-configuration.md
```diff
@@ -0,0 +1,833 @@
...
+
+---
+
+## Step 1: Create a Certificate Authority (CA)
+
+The Certificate Authority is used to sign both the server and client certificates, establishing trust between Docker and Galasa.
+
+### 1.1 Generate CA Private Key
+
+```bash
```
eamansour, line 41: The commands in this documentation are macOS/Linux-specific - are we able to get equivalent windows commands? Or perhaps mention that the instructions are only applicable for macOS/Linux users for now?

Happy for the windows commands to be added in a separate PR

docs/docker-tls-configuration.md
```diff
@@ -0,0 +1,833 @@
...
+  -out server-cert.pem \
+  -extfile extfile.cnf
+```
+
+<!-- ### 2.5 Set Proper Permissions
+
+```bash
+chmod 0400 ca-key.pem server-key.pem
+chmod 0444 ca.pem server-cert.pem
+``` -->
```
eamansour: Is this section needed? If not, could we remove it?

docs/docker-tls-configuration.md
```diff
@@ -0,0 +1,833 @@
...
+  -out cert.pem \
+  -extfile extfile-client.cnf
+```
+
+<!-- ### 3.5 Set Client Certificate Permissions
+
+```bash
+chmod 0400 key.pem
+chmod 0444 cert.pem
+``` -->
```
eamansour: Similar comment here regarding removing this commented-out section if it's not needed

docs/content/docs/managers/cloud-managers/docker-tls-configuration.md
```diff
@@ -0,0 +1,833 @@
...
+```
+
+**To get the base64 content:**
+```bash
+echo "base64:$(cat docker-tls-oneline.b64)"
+```
+
+#### Option B: ETCD Credentials Store
+
+For Galasa services, you can use the Secrets REST API to manage KeyStore credentials:
```
eamansour, line 739: While it's great to know that we can manage keystores using the REST API, should we also have an option in the `galasactl secrets set` command to be able to set keystores?

Something like:
```bash
galasactl secrets set --name DOCKER_TLS --keystore-file /path/to/keystore --keystore-password <password>
```

Happy for this to be done in a separate PR (potentially for a separate issue)

docs/docker-tls-configuration.md
```diff
@@ -0,0 +1,833 @@
...
+  - The KeyStore cannot be loaded
+  - The "base64:" prefix is missing
+
+**Secrets API Benefits:**
+- ✅ **Fail-fast validation:** Errors detected immediately at creation time, not during test execution
+- ✅ **Centralized management:** Single API for all credential types
+- ✅ **Audit trail:** Automatic tracking of who created/updated credentials and when
+- ✅ **Role-based access control (RBAC):** Fine-grained permissions for credential operations
+- ✅ **Encrypted storage:** All credentials encrypted at rest in etcd
+- ✅ **RESTful interface:** Easy integration with CI/CD pipelines and automation tools
```
eamansour: I don't think we need to list out the benefits of the secrets API here

docs/docker-tls-configuration.md
```diff
@@ -0,0 +1,833 @@
...
+
+### Using JKS with Galasa
+
+When using JKS format, you'll need to:
+
+1. **Base64 encode the JKS file** (see Step 6 below)
+2. **Set the KeyStore type to JKS** in Galasa configuration:
+
+```properties
+secure.credentials.DOCKER_TLS.keystore=base64:MIIKEQIBAzCCCdcGCSqGSIb3DQE...
```
eamansour: Could the encoding perhaps be a separate CPS property rather than a prefix for the value? Or perhaps we could omit the encoding completely if we only support base64, and document that the value must be in base64 format?

I can see people forgetting to put the prefix into the value when it comes to copy-pasting the encoded value.

docs/docker-tls-configuration.md
```diff
@@ -0,0 +1,833 @@
...
+public class DockerTlsTest {
+    
+    @DockerContainer(image = "alpine:latest", dockerEngineTag = "PRIMARY")
+    public IDockerContainer container;
+    
+    @Test
+    public void testDockerConnection() throws Exception {
+        // Container will be provisioned using TLS connection
+        String output = container.exec("echo", "Hello from secure Docker!");
+        System.out.println(output);
```
eamansour: Very minor but could we use a logger for this example instead of a println?

docs/docker-tls-configuration.md
```diff
@@ -0,0 +1,833 @@
...
+
+# Correct:
+secure.credentials.DOCKER_TLS.keystore=base64:MIIKEQIBAzCCCdcGCSqGSIb3DQE...
+```
+
+For Secrets REST API:
+```json
+{
+  "data": {
+    "keystore": "base64:MIIKEQIBAzCCCdcGCSqGSIb3DQE...",
```
eamansour: Perhaps the `base64` prefix here could be its own `encoding` field for the keystore?

modules/framework/galasa-parent/dev.galasa.framework.api.secrets/src/main/java/dev/galasa/framework/api/secrets/internal/routes/AbstractSecretsRoute.java
```diff
@@ -91,35 +99,106 @@ protected ICredentials buildDecodedCredentialsToSet(SecretRequest secretRequest,
...
+     * @param secretRequest the request containing the credential data
+     * @param keystore the keystore request object
+     * @return decoded keystore credentials
+     * @throws InternalServletException if decoding fails or keystore creation fails
+     */
+    private ICredentials decodeKeystoreCredentials(SecretRequest secretRequest, SecretRequestkeystore keystore) throws InternalServletException {
+        String decodedKeystore = decodeSecretValue(keystore.getvalue(), keystore.getencoding());
+        
+        SecretRequestKeystorePassword keystorePassword = secretRequest.getKeystorePassword();
+        String decodedKeystorePassword = decodeSecretValue(keystorePassword.getvalue(), keystorePassword.getencoding());
```
eamansour, line 194: Could `keystorePassword` be null here? If so, this could throw a null pointer exception

modules/framework/galasa-parent/dev.galasa.framework/src/test/java/dev/galasa/framework/internal/creds/CredentialsKeyStoreTest.java
```diff
@@ -0,0 +1,249 @@
...
+    private File fileCREDS;
+    private KeyStore testKeyStore;
+    private String testPassword = "testPassword123"; //pragma: allowlist secret
+    private byte[] testKeyStoreBytes;
+    private String encodedKeyStore;
+
+    @Before
+    public void setup() throws Exception {
+        fileCPS = File.createTempFile("galasacps_", ".properties");
+        fileCREDS = File.createTempFile("galasacreds_", ".properties");
```
eamansour: Instead of creating real files in a temp directory, we could use the existing MockFileSystem class

modules/framework/galasa-parent/dev.galasa.framework/src/test/java/dev/galasa/framework/internal/creds/CredentialsKeyStoreTest.java
```diff
@@ -0,0 +1,249 @@
...
+        keyStore.setEntry("test-secret-key", secretKeyEntry, protectionParam);
+
+        return keyStore;
+    }
+
+    private void saveProperties(Properties properties, File file) throws Exception {
+        FileOutputStream out = new FileOutputStream(file);
+        properties.store(out, null);
+        out.close();
+    }
```
eamansour: If we use a MockFileSystem, I don't believe we would need to do any IO operations to save properties to a file.

modules/managers/galasa-managers-parent/galasa-managers-cloud-parent/dev.galasa.docker.manager/src/main/java/dev/galasa/docker/internal/DockerEngineImpl.java
```diff
@@ -48,6 +57,7 @@ public class DockerEngineImpl implements IDockerEngine {
 	private final IDynamicStatusStoreService dss;
 
 	private URI uri;
+	private SSLContext sslContext;  // SSL context for HTTPS connections
```
eamansour: Minor - the inline comment here doesn't seem useful, so I'd be in favour of removing it

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/creds/CredentialsKeyStore.java
```diff
@@ -0,0 +1,191 @@
...
+     */
+    @Override
+    public KeyStore getKeyStore() {
+        if (this.keyStore == null) {
+            try {
+                this.keyStore = KeyStore.getInstance(this.keyStoreType);
+                this.keyStore.load(new ByteArrayInputStream(this.keyStoreBytes), this.keyStorePassword.toCharArray());
+            } catch (Exception e) {
+                throw new RuntimeException("Failed to load KeyStore of type " 
+                                            + this.keyStoreType + ": " + e.getMessage(), e);
```
eamansour: Instead of a RuntimeException, could we throw something like a CredentialsException?

Also, can the exception message/stack trace contain the keystore password in it? If so, we might need to avoid logging the stack trace here

eamansour (commented):
Overall looks good! I've added a few thoughts and suggestions to the PR - happy to approve once they're addressed!

modules/managers/galasa-managers-parent/galasa-managers-cloud-parent/dev.galasa.docker.manager/src/main/java/dev/galasa/docker/internal/DockerEngineImpl.java
jadecarino: Thanks Eamonn, agreed, removed in https://github.com/galasa-dev/galasa/pull/551/commits/517b9c210585bca93899bac998593aa90973352e

docs/content/docs/managers/cloud-managers/docker-tls-configuration.md
jadecarino: Thanks Eamonn, sure, I have added Windows commands to the documentation and generally improved them in this commit: https://github.com/galasa-dev/galasa/pull/551/commits/0bb710bb8298492d2711efb88e6c7e2d9496ae44

docs/docker-tls-configuration.md
jadecarino: Done in https://github.com/galasa-dev/galasa/pull/551/commits/0bb710bb8298492d2711efb88e6c7e2d9496ae44

docs/docker-tls-configuration.md
jadecarino: Done in https://github.com/galasa-dev/galasa/pull/551/commits/0bb710bb8298492d2711efb88e6c7e2d9496ae44

docs/content/docs/managers/cloud-managers/docker-tls-configuration.md
jadecarino: I think I will do this in a separate PR as this is a fairly large change set and I'd like to get it merged so we can switch over to using our protected Docker engine ASAP.

docs/docker-tls-configuration.md
jadecarino: Removed in https://github.com/galasa-dev/galasa/pull/551/commits/0bb710bb8298492d2711efb88e6c7e2d9496ae44

docs/docker-tls-configuration.md
jadecarino: Fixed in https://github.com/galasa-dev/galasa/pull/551/commits/0bb710bb8298492d2711efb88e6c7e2d9496ae44

modules/framework/galasa-parent/dev.galasa.framework/src/main/java/dev/galasa/framework/spi/creds/CredentialsKeyStore.java
jadecarino: Updates done in https://github.com/galasa-dev/galasa/pull/551/changes/84ef9f0c5dadf5ed2e9e060950ad8870bd669f7d

modules/framework/galasa-parent/dev.galasa.framework/src/test/java/dev/galasa/framework/internal/creds/CredentialsKeyStoreTest.java
jadecarino: I have updated these tests in https://github.com/galasa-dev/galasa/pull/551/commits/f2c010f1bd9c7006626db043f1a001636b727498  - didn't need to use the MockFileSystem actually but have removed the need for creating these temp files

modules/framework/galasa-parent/dev.galasa.framework/src/test/java/dev/galasa/framework/internal/creds/CredentialsKeyStoreTest.java
jadecarino: Removed the need for the IO operations in https://github.com/galasa-dev/galasa/pull/551/commits/f2c010f1bd9c7006626db043f1a001636b727498

docs/docker-tls-configuration.md
jadecarino: I have omitted the prefix completely to only support base64 for now, and have updated docs to reflect this. Changes are in https://github.com/galasa-dev/galasa/pull/551/commits/5ad89d9aa49684c6f136d3a56fd422259aed86e3

docs/docker-tls-configuration.md
jadecarino: Changes in https://github.com/galasa-dev/galasa/pull/551/commits/5ad89d9aa49684c6f136d3a56fd422259aed86e3

modules/framework/galasa-parent/dev.galasa.framework.api.secrets/src/main/java/dev/galasa/framework/api/secrets/internal/routes/AbstractSecretsRoute.java
jadecarino: Good point thanks, fixed in https://github.com/galasa-dev/galasa/pull/551/commits/1641dce0bafe4b7f9d41fe5950afa5314ea390f9

modules/framework/galasa-parent/dev.galasa/src/main/java/dev/galasa/ICredentialsKeyStore.java
```diff
@@ -38,7 +38,7 @@ public interface ICredentialsKeyStore extends ICredentials {
      *
      * @return KeyStore object containing certificates and keys
      */
-    KeyStore getKeyStore();
+    KeyStore getKeyStore() throws Exception;
```
eamansour, line 41: Minor - the CredentialsKeyStore implementation throws a `CredentialsException` which is great but this interface says it throws a generic `Exception` - can this be `CredentialsException` as well?

eamansour (approved):

modules/framework/galasa-parent/dev.galasa/src/main/java/dev/galasa/ICredentialsKeyStore.java
jadecarino: Unfortunately this isn't possible as this interface is in `dev.galasa`, I can't import in anything from `dev.galasa.framework` as that depends on `dev.galasa` so it gets into a loop and the Gradle build cannot work. This was the only way for build/compilation to succeed with the CredentialsKeyStore throwing CredentialsException. I did look through the `dev.galasa` bundle to check for any other examples of an exception being thrown and there weren't unfortunately.

eamansour: No problem, happy to leave as-is in that case.

PR #555, opened 2026-03-24
Iss2556 - Fix the `--user` flag on `runs submit` being case-sensitive

jadecarino:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2556.

CouchdbAuthStore falls back to a case-insensitive search on all users if the first exact search comes back with no users. This means that the `--user` flag is now case insensitive, and does not return an error if automation changes the casing of the user to different than is stored in CouchDB

## Changes
- [x] Change to CouchdbAuthStore to emit loginIds as lower-case to support case-insensitive search.
- [x] Updates to CLI syntax help, README.md and openapi.yaml to doc that 'user' is now case-insensitive.

### DIFF
--- a/modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
+++ b/modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
@@ -230,7 +230,7 @@ public class CouchdbAuthStore extends CouchdbStore implements IAuthStore {
         IUser user = null;
 
         try {
-            // Fetch documents matching the loginId
+            // Fetch documents matching the loginId (case-sensitive)
             userDocument = getAllDocsByLoginId(USERS_DATABASE_NAME, loginId, USERS_DB_VIEW_NAME);
 
             // Since loginIds are unique, there should be only one document.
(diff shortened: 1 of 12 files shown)


modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -248,6 +248,18 @@ public IUser getUserByLoginId(String loginId) throws AuthStoreException {
 
                 // Assign fetchedUser to the user variable
                 user = new UserImpl(fetchedUser);
+            } else {
+                // If exact match fails, do case-insensitive search through all users
+                logger.info("Exact match failed, performing case-insensitive search for loginId: " + loginId);
+                List<IUser> allUsers = getAllUsers();
```
eamansour, line 254: If there are thousands of user records, this could be a really slow search. Is there a way where we could change the Mango query to CouchDB to be case-insensitive, rather than having to get all users and search through them?

Or perhaps we should change how user records are created so when the records are made, we convert the user IDs to lower case and then always search based on lower case values?

jadecarino: Good catch thanks Eamonn - I've now removed this call to getAllUsers which could have caused performance issues and tried a new approach.

PR #557, opened 2026-03-30
Fix: Enable Docker Manager to authenticate with ICR

lewisja1406:
This fix addresses the issue where Galasa's Docker Manager cannot pull images from IBM Container Registry (ICR) due to authentication requirements.

Problem:
- ICR requires credentials to be sent to the OAuth token endpoint on the first request
- The bearer token alone is insufficient - Docker Engine requires base64-encoded JSON with username/password in the X-Registry-Auth header

Solution:
1. Modified retrieveBearerToken() to send Basic Auth credentials to the token endpoint on the first attempt (lines 170-185)
2. After successfully retrieving the bearer token, generate the proper Docker registry auth structure (base64-encoded JSON) for the X-Registry-Auth header (lines 195-203)

This enables Galasa to:
- Authenticate with ICR's token endpoint using Basic Auth
- Pull private images from ICR registries
- Maintain backward compatibility with registries that don't require credentials

Tested with:
- IBM Container Registry (uk.icr.io)
- Successfully pulled private images
- All integration tests passed

### DIFF
--- a/modules/managers/galasa-managers-parent/galasa-managers-cloud-parent/dev.galasa.docker.manager/src/main/java/dev/galasa/docker/internal/DockerRegistryImpl.java
+++ b/modules/managers/galasa-managers-parent/galasa-managers-cloud-parent/dev.galasa.docker.manager/src/main/java/dev/galasa/docker/internal/DockerRegistryImpl.java
@@ -149,36 +149,83 @@ public class DockerRegistryImpl {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


jadecarino (approved):
Changes look good, thanks for making this contribution Lewis! It looks as though the DCO check failed though. Do you have your signing keys set up for GitHub and if so would you mind signing your commit with `-s` please?

jadecarino:
Thanks for fixing signing! I am merging on your behalf.

PR #558, opened 2026-03-31
Add initial macOS credentials store implementation

eamansour:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2558 

This PR adds a new implementation of the credentials store that uses the macOS keychain to read credentials. The implementation uses the `security` CLI tool that is built in to macOS by default.

**Note**: There is a performance overhead of having to spawn a new process to run the `security` CLI tool to get credentials.

The alternative would be to call the macOS APIs directly, but this would make the code much more complex (using Java Native Access) and would mean we'd have to keep up-to-date with the macOS APIs.

If performance becomes an issue (e.g. thousands of credential reads per second which is unlikely), then we could investigate this further.

## Changes
- [x] Added an implementation of the macOS credentials store that has read-only access to the macOS keychain
- [x] Unit tests
- [x] Documentation updates
- [x] Release notes updates


modules/extensions/galasa-extensions-parent/dev.galasa.creds.os/src/main/java/dev/galasa/creds/os/internal/macos/SystemCommandExecutor.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Real implementation of CommandExecutor that executes system commands.
+ */
+public class SystemCommandExecutor implements CommandExecutor {
+
+    @Override
+    public CommandResult execute(String... command) throws OsCredentialsException {
+        try {
+            ProcessBuilder pb = new ProcessBuilder(command);
```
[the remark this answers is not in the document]
eamansour: Addressed in https://github.com/galasa-dev/galasa/pull/558/commits/c69b9d5cdb20159fe2d9938cf5ba97d52c30e7c5

modules/extensions/galasa-extensions-parent/dev.galasa.creds.os/src/main/java/dev/galasa/creds/os/internal/parsers/KeyStoreJsonParser.java
```diff
@@ -0,0 +1,66 @@
...
+    public ICredentials parse(JsonObject json) throws CredentialsException {
+        // Validate required fields
+        if (!json.has(FIELD_PASSWORD)) {
+            throw new OsCredentialsException(
+                "JSON KeyStore credential missing '" + FIELD_PASSWORD + "' field");
+        }
+        if (!json.has(FIELD_TYPE)) {
+            throw new OsCredentialsException(
+                "JSON KeyStore credential missing '" + FIELD_TYPE + "' field");
+        }
```
jadecarino: The current solution for KeyStore credentials means that users don't have to supply the CPS property `secure.credentials.MYKEYSTORE.type` as the CredentialsKeyStore object defaults it to `PKCS12` if not provided. Should we mirror that behaviour here? We could validate that if `json.has(FIELD_TYPE)` that it is one of the supported types, PKCS12 or JKS, instead of throwing an exception if it wasn't set at all.

modules/extensions/galasa-extensions-parent/dev.galasa.creds.os/src/test/java/dev/galasa/creds/os/internal/macos/MacOsKeychainStoreTest.java
```diff
@@ -0,0 +1,487 @@
...
+        json.addProperty("keystore", "abc");
+        json.addProperty("password", "pass");
+        String jsonPassword = gson.toJson(json);
+        mockCommandExecutor.setServiceName(serviceName);
+        mockCommandExecutor.addPassword(serviceName, "JSON", jsonPassword);
+
+        // When/Then
+        assertThatThrownBy(() -> store.getCredentials(credsId))
+            .isInstanceOf(OsCredentialsException.class)
+            .hasMessageContaining("JSON KeyStore credential missing 'type' field");
```
jadecarino: Related to above comment

jadecarino (approved):
Looks great! I wondered if we wanted to keep the same behaviour between file based and MacOS credentials for KeyStores in that "type" is an optional field as it defaults. Up to you, I don't have a strong inclination.

modules/extensions/galasa-extensions-parent/dev.galasa.creds.os/src/main/java/dev/galasa/creds/os/internal/parsers/KeyStoreJsonParser.java
eamansour: Thanks, I agree that behaviour should be here as well. Added in https://github.com/galasa-dev/galasa/pull/558/commits/30b135bdd0975229812602037386467320656897

modules/extensions/galasa-extensions-parent/dev.galasa.creds.os/src/test/java/dev/galasa/creds/os/internal/macos/MacOsKeychainStoreTest.java
eamansour: Added in https://github.com/galasa-dev/galasa/pull/558/commits/30b135bdd0975229812602037386467320656897

jadecarino (approved):

PR #562, opened 2026-04-02
Iss2556 - Fix `--user` option on `runs submit` being case-sensitive

### LINKED ISSUE #2556: The `--user` option is case-sensitive
## Describe the bug
The recent `--user` feature appears to be case-sensitive, which is troublesome for us.

We use Jenkins to fire off jobs that obtains the user ID firing off the job, which is then passed to the `--user` field. The trouble is, it appears the Jenkins instance we're using creates a users profile based off their OIDC provided email, all converted to lower-case.

Therefore the correct casing is lost, so when we fire off jobs via our FID (who has `owner` role) as a user, who may exist as a `tester`, but with incorrect casing, they don't match, resulting in an error stating `the users role doesn't have the `TEST_RUN_LAUNCH` action`.

I've manually tested the same command again but with the correct casing for `--user` and it works as expected.

jadecarino:
## Why?

Fixes https://github.com/galasa-dev/projectmanagement/issues/2556

The `--user` parameter for run submission was case-sensitive, causing permission errors when Jenkins passed user IDs in different casing than stored in the auth store.

This fix adds a case-insensitive user lookup method to the auth store interfaces and implements it using a new CouchDB view that emits lowercase loginIds. When submitting runs with `--user`, the system now performs case-insensitive matching against the auth store and uses the case-accurate loginId for permissions validation, and for storage in the Test Structure. Users can now submit runs with `--user` in any casing (e.g., admin, ADMIN, Admin) and it will correctly match the user in the auth store.

**Note:** This change set does not make searching by user with `--user` in `runs get` or with the REST API case-insensitive also. Currently, the Web UI provides dropdown options for users by retrieving them from CouchDB and these are in the correct case as stored in the auth store, so this is not an issue. For CLI and REST API calls, for now the `--user` field is still case sensitive. This could be completed as part of a second change set if requested.

## Changes
- [x] IAuthStore.java / IAuthStoreService.java: Added getUserByLoginIdCaseInsensitive() method for case-insensitive user lookup
- [x] AuthStoreDBLoginLowerCaseView.java: New CouchDB view bean for lowercase loginId matching
- [x] AuthStoreDBViews.java: Added reference to new lowercase view
- [x] CouchdbAuthStoreValidator.java: Added view definition that emits lowercase loginIds
- [x] CouchdbAuthStore.java: Implemented getUserByLoginIdCaseInsensitive() using the new CouchDB view
- [x] GroupRunsRoute.java: Updated run submission to use case-insensitive lookup and return case-accurate loginId for test structure storage
- [x] TestGroupRunsRoute.java: Added test for case-insensitive user lookup during run submission
- [x] MockAuthStoreService.java / MockAuthStore.java: Added mock implementation for testing
- [x] FrameworkAuthStoreService.java: Implemented delegation to underlying auth store
- [x] Tested on Minikube that ADMIN, admin and Admin could submit tests in the same way
- [x] Tested on Minikube that searching has not been regressed (although not yet case-insensitive)

### DIFF
--- a/modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
+++ b/modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
@@ -65,6 +65,7 @@ public class CouchdbAuthStore extends CouchdbStore implements IAuthStore {
 
     public static final String TOKENS_DB_VIEW_NAME = "loginId-view";
     public static final String USERS_DB_VIEW_NAME = "loginId-view";
+    public static final String USERS_DB_LOWERCASE_VIEW_NAME = "loginId-lowercase-view";
 
     private Log logger;
     private ITimeService timeService;
@@ -251,6 +252,47 @@ public class CouchdbAuthStore extends CouchdbStore implements IAuthStore {
             }
 
             logger.info("User retrieved from CouchDB OK");
+
+        } catch (CouchdbException e) {
+            String errorMessage = ERROR_FAILED_TO_RETRIEVE_USERS.getMessage(e.getMessage());
+            throw new AuthStoreException(errorMessage, e);
+        }
+
+        return user;
+    }
+
+    @Override
+    public IUser getUserByLoginIdCaseInsensitive(String loginId) throws AuthStoreException {
+        logger.info("Retrieving user by loginId (case-insensitive) from CouchDB");
+        List<ViewRow> userDocument;
+        IUser user = null;
+
+        try {
+            // Convert the loginId to lowercase for the lookup
+            String lowerCaseLoginId = loginId.toLowerCase();
+            
+            // Fetch documents matching the lowercase loginId using the lowercase view
+            userDocument = getAllDocsByLoginId(USERS_DATABASE_NAME, lowerCaseLoginId, USERS_DB_LOWERCASE_VIEW_NAME);
+
+            // Since loginIds are unique (case-insensitive), there should be only one document.
+            if (!userDocument.isEmpty()) {
+                ViewRow row = userDocument.get(0); // Get the first entry since loginId is unique
+
+                // Fetch the user document from the CouchDB using the ID from the row
+                UserDoc fetchedUser = getUserFromDocument(row.id);
+
+                if (row.value != null) {
+                    AuthDBNameViewDesign nameViewDesign = gson.fromJson(gson.toJson(row.value),
+                            AuthDBNameViewDesign.class);
+                    fetchedUser.setVersion(nameViewDesign._rev); // Set the version from the CouchDB rev
+                }
+
+                // Assign fetchedUser to the user variable
+                // The UserImpl will contain the actual case-accurate loginId from the database
+                user = new UserImpl(fetchedUser);
+            }
+
+            logger.info("User retrieved from CouchDB OK (case-insensitive lookup)");
 
         } catch (CouchdbException e) {
             String errorMessage = ERROR_FAILED_TO_RETRIEVE_USERS.getMessage(e.getMessage());
--- a/modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
+++ b/modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
@@ -235,7 +235,7 @@ public class CouchdbAuthStore extends CouchdbStore implements IAuthStore {
             userDocument = getAllDocsByLoginId(USERS_DATABASE_NAME, loginId, USERS_DB_VIEW_NAME);
 
             // Since loginIds are unique, there should be only one document.
-            if (!userDocument.isEmpty()) {
+            if (userDocument != null && !userDocument.isEmpty()) {
                 ViewRow row = userDocument.get(0); // Get the first entry since loginId is unique
 
                 // Fetch the user document from the CouchDB using the ID from the row
@@ -275,7 +275,7 @@ public class CouchdbAuthStore extends CouchdbStore implements IAuthStore {
             userDocument = getAllDocsByLoginId(USERS_DATABASE_NAME, lowerCaseLoginId, USERS_DB_LOWERCASE_VIEW_NAME);
 
             // Since loginIds are unique (case-insensitive), there should be only one document.
-            if (!userDocument.isEmpty()) {
+            if (userDocument != null && !userDocument.isEmpty()) {
                 ViewRow row = userDocument.get(0); // Get the first entry since loginId is unique
 
                 // Fetch the user document from the CouchDB using the ID from the row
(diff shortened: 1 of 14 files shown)


modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -260,6 +261,47 @@ public IUser getUserByLoginId(String loginId) throws AuthStoreException {
         return user;
     }
 
+    @Override
+    public IUser getUserByLoginIdCaseInsensitive(String loginId) throws AuthStoreException {
```
eamansour, line 265: Wouldn't it be better if we applied case-insensitive user searching to all operations rather than creating a new function here? That way, we maintain consistency in how the user-related searching operations work.

For example, if I run a `galasactl runs submit --user`, and I pass in my username in lowercase, that will now work. However, if I want to run a `galasactl users get --name`, it would fail to return my user.

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -260,6 +261,47 @@ public IUser getUserByLoginId(String loginId) throws AuthStoreException {
...
+
+        try {
+            // Convert the loginId to lowercase for the lookup
+            String lowerCaseLoginId = loginId.toLowerCase();
+            
+            // Fetch documents matching the lowercase loginId using the lowercase view
+            userDocument = getAllDocsByLoginId(USERS_DATABASE_NAME, lowerCaseLoginId, USERS_DB_LOWERCASE_VIEW_NAME);
+
+            // Since loginIds are unique (case-insensitive), there should be only one document.
+            if (!userDocument.isEmpty()) {
```
eamansour: Could userDocument be null here? It could throw a null pointer exception if so

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStoreValidator.java
```diff
@@ -44,6 +44,7 @@ public class CouchdbAuthStoreValidator extends CouchdbBaseValidator {
     // A couchDB view, it gets all the access tokens of a the user based on the loginId provided.
     public static final String DB_TABLE_TOKENS_DESIGN = "function (doc) { if (doc.owner && doc.owner.loginId) {emit(doc.owner.loginId, doc); } }";
     public static final String DB_TABLE_USERS_DESIGN = "function (doc) { if (doc['login-id']) { emit(doc['login-id'], doc); } }";
+    public static final String DB_TABLE_USERS_LOWERCASE_DESIGN = "function (doc) { if (doc['login-id']) { emit(doc['login-id'].toLowerCase(), doc); } }";
```
eamansour, line 47: We should avoid emitting the document as a value since it uses up a lot more disk space for no real reason (not sure why the other views above do this!)

Instead, if we want to get the user document in the response to the view query, we can pass in the `include_docs=true` query parameter in the request.

See https://docs.couchdb.org/en/stable/ddocs/views/intro.html:

> Do not emit the entire document as the value of your emit(key, value) statement unless you’re sure you know you want it. 
>
> This stores an entire additional copy of your document in the view’s secondary index. Views with emit(key, doc) take longer to update, longer to write to disk, and consume significantly more disk space. The only advantage is that they are faster to query than using the ?include_docs=true parameter when querying a view.

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
eamansour, line 265: Can we add unit tests for this method?

modules/framework/galasa-parent/dev.galasa.framework.api.runs/src/main/java/dev/galasa/framework/api/runs/routes/GroupRunsRoute.java
```diff
@@ -89,7 +92,16 @@ public HttpServletResponse handlePostRequest(
...
+                // Try to get the case-accurate loginId using case-insensitive lookup
+                String caseAccurateUser = getCaseAccurateLoginId(user);
+                // If we found a case-accurate user, use it; otherwise use the provided user
+                String userToValidate = (caseAccurateUser != null) ? caseAccurateUser : user;
+                
+                // Validate the user has permission to launch tests
+                validateActionPermitted(BuiltInAction.TEST_RUN_LAUNCH, userToValidate);
+                
+                // Update the schedule request with the validated user (case-accurate if found)
+                scheduleRequest.setUser(userToValidate);
```
eamansour, line 104: Do we have unit tests for these changes?

eamansour:
Looks good overall, just a few comments!

Regarding the new method to perform case-insensitive user searches, I think that should be the default behaviour for all user searches, but I'm aware that would affect several operations, so I'd be happy for that to be done in a separate PR.

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStoreValidator.java
eamansour: See `getDocumentsFromDatabaseViewByKey()` in the [CouchdbStore](https://github.com/galasa-dev/galasa/blob/main/modules/extensions/galasa-extensions-parent/dev.galasa.extensions.common.couchdb/src/main/java/dev/galasa/extensions/common/couchdb/CouchdbStore.java#L170) class which seems to do what you want it to do but has a parameter for `include_docs`

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
jadecarino: Good catch thanks Eamonn. I'll deliver this in a separate PR when I'm back as you've mentioned.

modules/framework/galasa-parent/dev.galasa.framework.api.runs/src/main/java/dev/galasa/framework/api/runs/routes/GroupRunsRoute.java
jadecarino: Good point thank you I have added some in https://github.com/galasa-dev/galasa/pull/562/commits/59371dc7149faf7c202149134cb8a3e82f32dbc1

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
jadecarino: Good point thank you I have added some in https://github.com/galasa-dev/galasa/pull/562/commits/59371dc7149faf7c202149134cb8a3e82f32dbc1

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
jadecarino: Good spot, fixed in https://github.com/galasa-dev/galasa/pull/562/commits/1112450cea8ee933fa5a2c45e6ffd3032acc3004

eamansour (approved):
Approved - as discussed, we'll sort out the issue with the views emitting entire documents as values in a separate PR

PR #570, opened 2026-04-13
James c iss2029 access token expiry

James-Cocker:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2029. 

This PR will add framework, REST API, CLI and docs changes to work towards the above issue of causing tokens to expire. We've made the decision to allow tokens to expire up to 1 year, with the default being 3 months. 

## Changes
- [x] Functionality


modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -27,10 +28,11 @@
 import dev.galasa.auth.couchdb.internal.beans.FrontEndClient;
 import dev.galasa.auth.couchdb.internal.beans.UserDoc;
 import dev.galasa.extensions.common.api.HttpClientFactory;
-import dev.galasa.extensions.common.api.LogFactory;
+import org.apache.commons.logging.LogFactory;
```
James-Cocker: Note: Need to change this back to dev.galasa.extensions.common.api.LogFactory

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -67,21 +69,147 @@ public class CouchdbAuthStore extends CouchdbStore implements IAuthStore {
     public static final String USERS_DB_VIEW_NAME = "loginId-view";
     public static final String USERS_DB_LOWERCASE_VIEW_NAME = "loginId-lowercase-view";
 
-    private Log logger;
+    // Default token lifespan in days (90 days)
+    public static final int DEFAULT_TOKEN_LIFESPAN_DAYS = 90;
```
James-Cocker, line 75: Note: this needs to take in the token lifespan instead of hardcoding one.

docs/content/docs/ecosystem/ecosystem-authentication.md
```diff
@@ -26,6 +28,8 @@ The `galasactl` tool will login implicitly when it needs to contact the remote G
 
 On a successful login using the `galasactl` tool, the `GALASA_TOKEN` will be used to create a new temporary bearer token, which is stored in the `bearer-tokens` folder in the Galasa home directory. This file contains a bearer token that galasactl uses to authenticate requests when communicating with a Galasa Ecosystem. If the bearer token expires, galasactl automatically attempts to re-authenticate with the Galasa Ecosystem using the configured `GALASA_TOKEN`.
 
+If your personal access token is approaching its expiry date, the CLI will display a warning message (GAL2001W) when you use it to authenticate. This gives you time to create a new token before the current one expires. Once a token expires, it can no longer be used to exchange for a JWT and becomes unusable. Note: JWTs last for 24 hours, so you might find your token will still work for a short while after it expires.
```
eamansour: If the Galasa service admin hasn't changed the JWT expiry time, then JWTs will be valid for 24 hours. If they have, then this note will be incorrect.

It's probably worth editing this note to say that JWTs may continue to be valid for a period of time determined by your Galasa service administrator, or removing the note entirely.

```
Note: JWTs last for 24 hours, so you might find your token will still work for a short while after it expires.
```

docs/content/docs/ecosystem/ecosystem-authentication.md
```diff
@@ -73,7 +77,7 @@ id: m.smith@gmail.com
 
 ### Revoking personal access tokens
 
-If a user moves to a new role, or loses a laptop with their personal access token on it, you can log into the Galasa Ecosystem and revoke their access token by using the `galasactl auth tokens delete` command. This ensures that the Ecosystem and systems-under-test remain secure.
+If a user moves to a new role, or loses a laptop with their personal access token on it, you can log into the Galasa Ecosystem and revoke their access token by using the `galasactl auth tokens delete` command. This ensures that the Ecosystem and systems-under-test remain secure. Tokens are also automatically disabled when they reach their expiry date.
```
eamansour: Instead of `Tokens are also automatically disabled`, maybe it should be `Tokens are also automatically revoked` since `disabled` technically means tokens are still valid?

docs/content/docs/ecosystem/ecosystem-authentication.md
eamansour: I think we can remove the warning message ID (the `GAL2001W` bit) since the ID means nothing to the user.

docs/content/docs/ecosystem/ecosystem-installing-k8s.md
```diff
@@ -277,6 +277,14 @@ Dex supports many connectors including Microsoft, LDAP, OIDC, and more. See the
 
 Update the `expiry` section to configure the expiry of JSON Web Tokens (JWTs) and refresh tokens issued by Dex. By default, JWTs expire 24 hours after being issued and refresh tokens remain valid unless they have not been used for one year. See the [Dex documentation on ID tokens](https://dexidp.io/docs/tokens){target="_blank"} for information and available expiry settings.
 
+#### Optional: Configure Personal Access Token Expiry Warning
+
+You can configure when users receive warnings about their personal access tokens approaching expiry by setting the CPS property `service.tokens.lifespan.nearly.expired.warning.days`. This property defines how many days before expiry the CLI will display a warning message (GAL2001W) when the token is used.
```
eamansour: Similar comment here - I think we can remove the warning message ID (the `GAL2001W` bit) since the ID means nothing to the user.

docs/content/releases/posts/v0.48.0.md
```diff
@@ -9,6 +9,16 @@ links:
 
 ## Changes affecting the Galasa Service
 
+- **Personal Access Token Expiry**: Personal access tokens now have expiry times associated with them. When a token expires, it can no longer be used to authenticate with the Galasa service.
+  - All existing tokens will migrate to expire in 90 days from the date of the update, however, they might expire sooner due to dex (see your helm chart).
```
eamansour: `they might expire sooner due to dex (see your helm chart)` could become `they might expire sooner depending on your Galasa service's configured Dex 'expiry' values in the Galasa service's Helm chart` or similar?

Feel free to experiment, I just think that `due to dex (see your helm chart)` should be clarified.

docs/content/releases/posts/v0.48.0.md
```diff
@@ -9,6 +9,16 @@ links:
...
+  - The Galasa webui will now present you with the option to set an expiry date for your token, and display if tokens are expired or nearing expiry.
+  - Tokens are automatically disabled when they reach their expiry date. Note: if you create a JWT close to the expiry date, then you might notice that your token will continue to work up to 24 hours after the expiry date, this is because the token is still valid in the JWT library.
```
eamansour: See previous comment about JWTs being valid for 24 hours (this is only true if you haven't changed the default JWT expiry values in the Helm chart).

Also, `this is because the token is still valid in the JWT library` isn't quite accurate since there's no `JWT library`. It could say `this is because the JWT has not expired yet` or similar?

docs/content/releases/posts/v0.48.0.md
```diff
@@ -9,6 +9,16 @@ links:
...
+  - The Galasa CLI will automatically delete your token if you attempt to use it after it has expired.
+  - The CLI displays a warning (GAL2001W) when using a token that is approaching its expiry date.
```
eamansour: Remove warning ID here too

modules/cli/pkg/auth/authenticator.go
```diff
@@ -156,8 +163,40 @@ func (authenticator *authenticatorImpl) getJwtFromRestApi(apiServerUrl string, a
 		} else {
 			log.Println("Bearer token received from API server OK")
 			jwt = tokenResponse.GetJwt()
+
+			// Check if the token is approaching expiry and display a warning
+			if tokenResponse.HasTokenExpiryTime() {
+				warningDays := 14 // Default value
```
eamansour: Can we put the default value into a `const` (to avoid magic numbers) - we then won't need the `// Default value` comment

modules/cli/pkg/auth/authenticator.go
```diff
@@ -156,8 +163,40 @@ func (authenticator *authenticatorImpl) getJwtFromRestApi(apiServerUrl string, a
...
+	expiryTime, err := time.Parse(time.RFC3339, expiryTimeStr)
+	if err != nil {
+		log.Printf("Failed to parse token expiry time: %v", err)
+		return
+	}
+
+	now := authenticator.timeService.Now()
+	daysUntilExpiry := expiryTime.Sub(now).Hours() / 24
+
+	if daysUntilExpiry <= float64(warningThresholdDays) && daysUntilExpiry > 0 {
```
eamansour: Why is `warningThresholdDays` being converted to a float64 here?  Would be great to keep days as whole numbers.

modules/cli/pkg/cmd/authLogin.go
```diff
@@ -127,6 +127,13 @@ func (cmd *AuthLoginComamnd) executeAuthLogin(
 
 		if err == nil {
 			authenticator := commsClient.GetAuthenticator()
+
+			// Set the console on the authenticator so it can display warnings
+			console := factory.GetStdOutConsole()
+			if authImpl, ok := authenticator.(interface{ SetConsole(spi.Console) }); ok {
+				authImpl.SetConsole(console)
+			}
```
eamansour: If the authenticator needs a console, can we please set the console when the authenticator is constructed rather than doing that here. We can then remove the SetConsole method if we do that and keep this code simple.

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -68,6 +70,10 @@ public class CouchdbAuthStore extends CouchdbStore implements IAuthStore {
     public static final String USERS_DB_LOWERCASE_VIEW_NAME = "loginId-lowercase-view";
 
     private Log logger;
+
+    // Default token lifespan in days (90 days)
```
eamansour: Can we remove this comment - it adds no value since the constant is descriptive

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -82,6 +88,138 @@ public CouchdbAuthStore(
...
+
+        // For each user, get their tokens
+        for (IUser user : users) {
+            String loginId = user.getLoginId();
+
+            List<IInternalAuthToken> tokens = null;
+            try {
+                tokens = getTokensByLoginId(loginId);
+            } catch (Exception e) {
+                logger.warn("Failed to retrieve tokens for user " + loginId + ", skipping user and continuing with others: " + e.getMessage());
```
eamansour: We don't want to print out the login ID since it could leak personal information (e.g. if the IDs are email addresses)

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -82,6 +88,138 @@ public CouchdbAuthStore(
...
+                continue;
+            }
+
+            for (IInternalAuthToken token : tokens) {
+                tokensProcessed++;
+
+                if (token.getExpiryTime() == null) {
+                    logger.info("Token " + token.getTokenId() + " for user " + loginId
```
eamansour: We don't want to print out the login ID since it could leak personal information (e.g. if the IDs are email addresses) - I'd remove this message entirely.

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -82,6 +88,138 @@ public CouchdbAuthStore(
...
+                    // Calculate expiry time: current time + constant amount of days
+                    Instant newExpiryTime = timeService.now().plus(DEFAULT_TOKEN_LIFESPAN_DAYS, ChronoUnit.DAYS);
+                    logger.info("Setting expiry time to: " + newExpiryTime);
+
+                    // Update the token with the new expiry time
+                    try {
+                        updateTokenWithExpiryTime(token.getTokenId(), newExpiryTime);
+                    } catch (AuthStoreException e) {
+                        tokensFailed++;
+                        logger.warn("Failed to migrate token " + token.getTokenId() + " for user " + loginId
```
eamansour: We don't want to print out the login ID since it could leak personal information (e.g. if the IDs are email addresses)

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -82,6 +88,138 @@ public CouchdbAuthStore(
...
+     * the migration process and is not exposed via the IAuthStore interface.
+     */
+    private void updateTokenWithExpiryTime(String tokenId, Instant expiryTime) throws AuthStoreException {
+        try {
+            // Get the existing token document
+            CouchdbAuthToken existingToken = getDocumentFromDatabase(TOKENS_DATABASE_NAME, tokenId,
+                    CouchdbAuthToken.class);
+
+            if (existingToken == null) {
+                String errorMessage = ERROR_FAILED_TO_RETRIEVE_TOKENS.getMessage("Token not found: " + tokenId);
```
eamansour: Maybe this error should be `ERROR_TOKEN_NOT_FOUND` instead of `ERROR_FAILED_TO_RETRIEVE_TOKENS`?

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -178,6 +319,33 @@ private IInternalAuthToken getAuthTokenFromDocument(String documentId) throws Co
...
+        logger.info("Retrieving token by Dex client ID from CouchDB");
+
+        try {
+            // Get all tokens and search for one with matching clientId
+            List<IInternalAuthToken> allTokens = getTokens();
+
+            for (IInternalAuthToken token : allTokens) {
+                if (token instanceof CouchdbAuthToken) {
+                    CouchdbAuthToken couchdbToken = (CouchdbAuthToken) token;
+                    if (clientId.equals(couchdbToken.getDexClientId())) {
```
eamansour: The `IInternalAuthToken` interface already has a `getDexClientId()` method, is there a reason why we're converting the token into a CouchdbAuthToken here?

If not, we could remove the cast and simply use `token.getDexClientId()`

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -178,6 +319,33 @@ private IInternalAuthToken getAuthTokenFromDocument(String documentId) throws Co
         return getDocumentFromDatabase(TOKENS_DATABASE_NAME, documentId, CouchdbAuthToken.class);
     }
 
+    @Override
+    public IInternalAuthToken getTokenByDexClientId(String clientId) throws AuthStoreException {
+        logger.info("Retrieving token by Dex client ID from CouchDB");
+
+        try {
+            // Get all tokens and search for one with matching clientId
+            List<IInternalAuthToken> allTokens = getTokens();
```
eamansour, line 329: This could be an expensive operation if there are lots of users with lots of tokens. We would likely need a new tokens view that uses client IDs as the keys that we could then search directly.

This would make the PR even bigger and we aren't having any performance issues so happy to keep as-is for now.

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -178,6 +319,33 @@ private IInternalAuthToken getAuthTokenFromDocument(String documentId) throws Co
...
+
+            for (IInternalAuthToken token : allTokens) {
+                if (token instanceof CouchdbAuthToken) {
+                    CouchdbAuthToken couchdbToken = (CouchdbAuthToken) token;
+                    if (clientId.equals(couchdbToken.getDexClientId())) {
+                        logger.info("Token found for Dex client ID: " + clientId);
```
eamansour: Let's avoid logging the clientId here since the clientId makes up part of a galasa token

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -178,6 +319,33 @@ private IInternalAuthToken getAuthTokenFromDocument(String documentId) throws Co
...
+                        return token;
+                    }
+                }
+            }
+
+            logger.info("No token found for Dex client ID: " + clientId);
```
eamansour: Let's avoid logging the clientId here since the clientId makes up part of a galasa token

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -178,6 +319,33 @@ private IInternalAuthToken getAuthTokenFromDocument(String documentId) throws Co
...
+        try {
+            // Get all tokens and search for one with matching clientId
+            List<IInternalAuthToken> allTokens = getTokens();
+
+            for (IInternalAuthToken token : allTokens) {
+                if (token instanceof CouchdbAuthToken) {
+                    CouchdbAuthToken couchdbToken = (CouchdbAuthToken) token;
+                    if (clientId.equals(couchdbToken.getDexClientId())) {
+                        logger.info("Token found for Dex client ID: " + clientId);
+                        return token;
```
eamansour: For consistency, can we have a single return statement at the end of this function?

modules/framework/galasa-parent/dev.galasa.framework.api.authentication/src/main/java/dev/galasa/framework/api/authentication/internal/routes/AuthRoute.java
```diff
@@ -246,7 +246,8 @@ private void addTokenToAuthStore(String clientId, String jwt, String description
         IInternalUser user = new InternalUser(jwtWrapper.getUsername(), jwtWrapper.getSubject());
 
         try {
-            authStoreService.storeToken(clientId, description, user);
+            // Default to 90 days for deprecated route
+            authStoreService.storeToken(clientId, description, user, 90);
```
eamansour: Can we please put the default `90` days in a constant to avoid having a magic number here? The comment won't be needed either as the constant should be self-explanatory.

modules/framework/galasa-parent/dev.galasa.framework.api.authentication/src/main/java/dev/galasa/framework/api/authentication/internal/routes/AuthTokensRoute.java
```diff
@@ -297,19 +333,25 @@ private JsonObject sendTokenPost(TokenPayload requestBodyJson)
...
      */
-    private void addTokenToAuthStore(String clientId, String jwt, String description) throws InternalServletException {
+    private void addTokenToAuthStore(String clientId, String jwt, String description, Integer tokenLifespanDays)
+            throws InternalServletException {
         logger.info("Storing new token record in the auth store");
         JwtWrapper jwtWrapper = new JwtWrapper(jwt, env);
         IInternalUser user = new InternalUser(jwtWrapper.getUsername(), jwtWrapper.getSubject());
 
+        // Default to 90 days if not provided
+        int lifespanDays = (tokenLifespanDays != null) ? tokenLifespanDays : 90;
```
eamansour: Another magic number here - can we use a constant for the default and remove the comment?

modules/framework/galasa-parent/dev.galasa.framework.api.authentication/src/main/java/dev/galasa/framework/api/authentication/internal/routes/AuthTokensRoute.java
```diff
@@ -229,6 +262,9 @@ public HttpServletResponse handlePostRequest(String pathInfo,
                 throw new InternalServletException(error, HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
             }
 
+        } catch (InternalServletException e) {
+            // Re-throw InternalServletException to preserve the specific error code
+            throw e;
```
eamansour: Is this catch needed? We could let the exception propagate out like it was previously

modules/framework/galasa-parent/dev.galasa.framework.api.authentication/src/main/java/dev/galasa/framework/api/authentication/internal/routes/AuthTokensRoute.java
```diff
@@ -60,12 +64,11 @@ public class AuthTokensRoute extends PublicRoute {
     private IOidcProvider oidcProvider;
     private IDexGrpcClient dexGrpcClient;
     private Environment env;
-
+    private IFramework framework;
```
eamansour: I don't think we need the framework stored as a field here. We can pull out the services we need from the framework in the route's constructor.

modules/framework/galasa-parent/dev.galasa.framework.api.authentication/src/main/java/dev/galasa/framework/api/authentication/internal/routes/AuthTokensRoute.java
```diff
@@ -328,15 +370,66 @@ private List<AuthToken> convertAuthStoreTokenIntoTokenBeans(List<IInternalAuthTo
...
+     * Returns the default value if the property is not set or invalid.
+     * Validates that the value is between MIN and MAX days.
+     *
+     * @return the number of days before expiry when a warning should be displayed
+     */
+    private int getTokenExpiryWarningDays() {
+        int warningDays = DEFAULT_TOKEN_EXPIRY_WARNING_DAYS;
+
+        try {
+            IConfigurationPropertyStoreService cps = framework.getConfigurationPropertyService("service");
```
eamansour: Can we move getting the CPS service into the constructor? This could be expensive doing it every time this route is called.

modules/framework/galasa-parent/dev.galasa.framework.api.authentication/src/main/java/dev/galasa/framework/api/authentication/internal/routes/AuthTokensRoute.java
```diff
@@ -328,15 +370,66 @@ private List<AuthToken> convertAuthStoreTokenIntoTokenBeans(List<IInternalAuthTo
...
+            // CPS property format: service.tokens.lifespan.nearly.expired.warning.days
+            String warningDaysStr = cps.getProperty("tokens.lifespan.nearly.expired.warning", "days");
```
eamansour: Could we put this property into constants?

modules/framework/galasa-parent/dev.galasa.framework.api.authentication/src/main/java/dev/galasa/framework/api/authentication/internal/routes/AuthTokensRoute.java
```diff
@@ -396,4 +490,39 @@ private boolean isLoggingIntoWebUI(String refreshToken, String tokenDescription)
...
+                    // Delete the expired token
+                    authService.revokeToken(token.getTokenId(), userId);
+
+                    ServletError error = new ServletError(GAL5454_TOKEN_EXPIRED);
+                    throw new InternalServletException(error, HttpServletResponse.SC_UNAUTHORIZED);
+                }
+            }
+        } catch (AuthStoreException e) {
+            ServletError error = new ServletError(GAL5000_GENERIC_API_ERROR);
+            throw new InternalServletException(error, HttpServletResponse.SC_INTERNAL_SERVER_ERROR, e);
```
eamansour: Can we use a more specific error message here, rather than the `GAL5000_GENERIC_API_ERROR`?

modules/framework/galasa-parent/dev.galasa.framework.api.authentication/src/main/java/dev/galasa/framework/api/authentication/internal/TokenPayloadValidator.java
```diff
@@ -24,11 +24,17 @@ public void validate(TokenPayload tokenPayload) throws InternalServletException
...
+            Integer tokenLifespanDays = tokenPayload.getTokenLifespanDays();
     
             if (clientId != null && isAlphanumWithDashes(clientId)) {
                 // The refresh token and authorization code parameters are mutually exclusive
                 isValid = ((refreshToken != null && isAlphanumWithDashes(refreshToken) && authCode == null)
                     || (authCode != null && isAlphanumWithDashes(authCode) && refreshToken == null));
+
+                // Validate token_lifespan_days if provided
+                if (isValid && tokenLifespanDays != null) {
+                    isValid = (tokenLifespanDays >= 1 && tokenLifespanDays <= 365);
```
eamansour: Can we put the min and max values for token lifespans in constants?

modules/cli/pkg/auth/authenticator.go
James-Cocker: I believe that was because `.Hours() / 24` produces a float, but i've casted it directly to an int now.

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
James-Cocker: Would `getUserNumber()` be appropriate instead?

eamansour: Yes that would be safer than the login ID

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
James-Cocker: Oops great point thank you!

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
James-Cocker: Agreed completely - it was a reluctant move from me at the time as well.

modules/framework/galasa-parent/dev.galasa.framework.api.authentication/src/main/java/dev/galasa/framework/api/authentication/internal/routes/AuthTokensRoute.java
James-Cocker: Yes you're completely right 😁

modules/framework/galasa-parent/dev.galasa.framework.api.authentication/src/main/java/dev/galasa/framework/api/authentication/internal/routes/AuthTokensRoute.java
James-Cocker: Does `GAL5053_FAILED_TO_RETRIEVE_TOKENS` work alright for you or should i consider making a new one? This would be for when `getTokenByDexClientId()` fails.

eamansour: Sure, that'll be better than the generic error :)

modules/framework/galasa-parent/dev.galasa.framework.api.authentication/src/main/java/dev/galasa/framework/api/authentication/internal/routes/AuthTokensRoute.java
```diff
@@ -396,4 +490,39 @@ private boolean isLoggingIntoWebUI(String refreshToken, String tokenDescription)
...
+
+                // Check if the token has expired
+                if (expiryTime != null && now.isAfter(expiryTime)) {
+                    logger.info("Personal access token has expired. Deleting token from auth store.");
+
+                    // Delete the expired token
+                    authService.revokeToken(token.getTokenId(), userId);
+
+                    ServletError error = new ServletError(GAL5454_TOKEN_EXPIRED);
+                    throw new InternalServletException(error, HttpServletResponse.SC_UNAUTHORIZED);
```
eamansour, line 523: The `POST /auth/tokens` entry in the openapi.yaml doc needs to mention that a 401 could be returned in the response. It currently lists these possible response codes:

```yaml
      responses:
        '200':
          description: Returns a JSON Object containing a JWT and refresh token
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/TokenResponse'
        '400':
          description: Bad Request
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/APIError'
              examples:
                missingparametererror:
                  value:
                    error_code: 5400
                    error_message: "E: Error occurred when trying to execute request '/auth/tokens'. Check your request parameters or report the problem to your Galasa Ecosystem owner."
                  summary: One or more required body properties are missing
```

I definitely think that exchanging refresh tokens + client IDs for a JWT should be a separate API endpoint since this endpoint is doing too much (`POST /auth/tokens` is also a public endpoint so the possibility of returning 401 seems odd), but that can be done in a separate issue!

modules/framework/galasa-parent/dev.galasa.framework.api.authentication/src/main/java/dev/galasa/framework/api/authentication/internal/routes/AuthTokensRoute.java
```diff
@@ -108,7 +110,9 @@ public AuthTokensRoute(
         this.env = env;
         this.rbacService = rbacService;
         this.timeService = timeService;
-        this.framework = framework;
+        this.configurationPropertyStoreService = (framework != null)
+                ? framework.getConfigurationPropertyService("service")
+                : null;
```
eamansour: The framework should never be null (otherwise the API server wouldn't start up properly), so we could simplify this to be:

```suggestion
this.configurationPropertyStoreService = framework.getConfigurationPropertyService("service");
```

modules/framework/galasa-parent/dev.galasa.framework.api.authentication/src/main/java/dev/galasa/framework/api/authentication/internal/routes/AuthTokensRoute.java
```diff
@@ -389,10 +387,17 @@ private List<AuthToken> convertAuthStoreTokenIntoTokenBeans(List<IInternalAuthTo
...
+                    + DEFAULT_TOKEN_EXPIRY_WARNING_DAYS + " days.");
+            return warningDays;
+        }
+
         try {
-            IConfigurationPropertyStoreService cps = framework.getConfigurationPropertyService("service");
             // CPS property format: service.tokens.lifespan.nearly.expired.warning.days
-            String warningDaysStr = cps.getProperty("tokens.lifespan.nearly.expired.warning", "days");
+            String warningDaysStr = configurationPropertyStoreService
+                    .getProperty("tokens.lifespan.nearly.expired.warning", "days");
```
eamansour: Minor - can this CPS property be in a constant?

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -113,6 +113,18 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/APIError'
+              examples:
+                tokenexpired:
+                  value:
+                    error_code: 5454
+                    error_message: "E: Your personal access token has expired. Create a new personal access token on your Galasa service's web user interface to continue to authenticate with the Galasa service."
+                  summary: The personal access token used for refresh has expired
```
eamansour: We can re-use the existing "Unauthorized" response like this:

```
        '401':
          $ref: "#/components/responses/Unauthorized"
```

eamansour (approved):

PR #573, opened 2026-04-14
Add istio sidecar to test pods if istio is enabled in the Galasa service

eamansour:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/1531. 
Relates to changes in https://github.com/galasa-dev/helm/pull/121

## Changes
- [x] Added galasa.dev docs on how to configure the Galasa service to use Istio and what's required in order to do so
- [x] Added the `is_istio_enabled` setting which is used to determine whether to inject an istio sidecar to test pods

### DIFF
--- a/modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/test/java/dev/galasa/framework/k8s/controller/TestPodSchedulerTest.java
+++ b/modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/test/java/dev/galasa/framework/k8s/controller/TestPodSchedulerTest.java
@@ -294,6 +294,45 @@ public class TestPodSchedulerTest {
     }
 
     @Test
+    public void testCanCreateTestPodWithIstioSidecarLabelWhenIstioEnabled() throws Exception {
+        // Given...
+        MockEnvironment mockEnvironment = new MockEnvironment();
+
+        String encryptionKeysMountPath = "/encryption/encryption-keys.yaml";
+        mockEnvironment.setenv(FrameworkEncryptionService.ENCRYPTION_KEYS_PATH_ENV, encryptionKeysMountPath);
+
+        MockIDynamicStatusStoreService mockDss = new MockIDynamicStatusStoreService();
+        MockFrameworkRuns mockFrameworkRuns = new MockFrameworkRuns(new ArrayList<>());
+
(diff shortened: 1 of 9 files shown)


modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/test/java/dev/galasa/framework/k8s/controller/TestPodSchedulerTest.java
```diff
@@ -250,6 +254,45 @@ public void testCanCreateTestPodOk() throws Exception {
         assertPodDetailsAreCorrect(pod, galasaServiceInstallName, runName, podName, expectedEncryptionKeysMountPath, settings);
     }
 
+    @Test
+    public void testCanCreateTestPodWithoutIstioSidecarLabelWhenIstioDisabled() throws Exception {
+        // Given...
```
jadecarino, line 259: Do we also need a unit test for testCanCreateTestPodWithIstioSidecarLabelWhenIstioEnabled?

jadecarino (approved):
Looks great! Just wondering if we should have an additional unit test for the other scenario in TestPodSchedulerTest

modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/test/java/dev/galasa/framework/k8s/controller/TestPodSchedulerTest.java
eamansour: By default, the `MockISettings()` sets the `isIstioEnabled` field to `true` so the label is added for the other tests but I agree that's not immediately obvious, so I've added an explicit test for this in https://github.com/galasa-dev/galasa/pull/573/commits/5ae33033f74b8c3a8a3045f3efa0cd00e5b54be6

jadecarino (approved):

modules/framework/galasa-parent/dev.galasa.framework.k8s.controller/src/test/java/dev/galasa/framework/k8s/controller/TestPodSchedulerTest.java
jadecarino: Great thanks!

PR #574, opened 2026-04-14
Operations that involving searching the Auth Store by user/loginId are now case insensitive

jadecarino:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2570

This change set builds upon changes made in #562 by changing the behaviour of all searches into the Auth Store by loginId to be case insensitive. This replaces the code in method getUserByLoginId with getUserByLoginIdCaseInsensitive and ensures all callers use this method. This change set also includes updates to the Tokens and User views from CouchDB. Both views now emit the loginId from CouchDB in lower-case, so the methods that call these views for searching can search case insensitively (the search string is lower-cased to match the lower-case emitted loginId, then the tokens/users are returned). Due to these changes, the view loginId-lowercase-view introduced in #562 has been deleted and all associated code.

## Changes
- [x] Unit tests - consolidated duplicate tests together now that case insensitive searching is the default behaviour (if applicable)


modules/cli/docs/generated/galasactl_users_delete.md
```diff
@@ -14,7 +14,7 @@ galasactl users delete [flags]
 
 ```
   -h, --help              Displays the options for the 'users delete' command.
-      --login-id string   A mandatory field indicating the login ID of a user.
+      --login-id string   A mandatory field indicating the login ID of a user.This parameter is case-insensitive (e.g., 'myuser', 'MyUser', and 'MYUSER' are treated as the same user).
```
eamansour: Minor - there's a space missing in `user.This`

The same applies to the other galasactl users commands

eamansour (approved):
A couple of minor formatting issues but overall looks good!

modules/cli/docs/generated/galasactl_users_delete.md
jadecarino: Great spot thanks, fixed in https://github.com/galasa-dev/galasa/pull/574/commits/fddead1dc4ddbdf75679dcfa29fcc76d6b85a452

PR #577, opened 2026-04-16
CouchdbAuthStoreValidator reapplies database views if missing or updated

jadecarino:
## Why?

Fixes an issue introduced after #574 which changed the tokens and users view. The CouchdbAuthStoreValidator previously only applied the view if it was missing, but now it applies it if it's different to expected (what is defined in the code). This should fix code changes to the view function not being picked up and used by CouchDB.

### DIFF
--- a/modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStoreValidator.java
+++ b/modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStoreValidator.java
@@ -131,14 +131,16 @@ public class CouchdbAuthStoreValidator extends CouchdbBaseValidator {
             tableDesign.views.loginIdView = new AuthStoreDBLoginView();
         }
 
-        if (tableDesign.views.loginIdView.map == null) {
+        String expectedMapFunction;
+        if (dbName.equals(CouchdbAuthStore.TOKENS_DATABASE_NAME)) {
+            expectedMapFunction = DB_TABLE_TOKENS_DESIGN;
+        } else {
+            expectedMapFunction = DB_TABLE_USERS_DESIGN;
+        }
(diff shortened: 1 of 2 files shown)


modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStoreValidator.java
```diff
@@ -131,14 +131,16 @@ protected boolean updateDesignDocToDesiredDesignDoc(AuthDBNameViewDesign tableDe
...
+
+        if (tableDesign.views.loginIdView.map == null || !tableDesign.views.loginIdView.map.equals(expectedMapFunction)) {
             isUpdated = true;
-            if(dbName.equals(CouchdbAuthStore.TOKENS_DATABASE_NAME)){
-                tableDesign.views.loginIdView.map = DB_TABLE_TOKENS_DESIGN;
-            }
-            else{
-                tableDesign.views.loginIdView.map = DB_TABLE_USERS_DESIGN;
-            }
+            tableDesign.views.loginIdView.map = expectedMapFunction;
```
eamansour, line 143: Changes look good - it would be great if we could also have some unit tests to check this works and guard against any regressions in the future

eamansour (approved):

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStoreValidator.java
jadecarino: Yes I should have added these! Added in https://github.com/galasa-dev/galasa/pull/577/commits/33dc4c07e4277efcd54ed9c5ea00575045b5a6e3

eamansour (approved):

PR #578, opened 2026-04-16
Allow query parameters with no values to be supplied in the HTTP manager

eamansour:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2575. 

The HTTP manager no longer requires query parameters to be in the form `key=value` and will now allow query parameters with no values to be supplied without throwing exceptions.

## Changes
- [x] Updated the `buildUri` method in the HTTP manager to use the URIBuilder to build request URIs rather than regex which also allows query parameters to be supplied without values
- [x] Updated the HttpManagerIVT's `buildURITest` test case assertions to expect valid responses from httpbin
  - Previously, the test believed `?should==fail` was invalid, but this is a valid query parameter which has the value `=fail` and similarly, it thought that `thisparam=ok&&oops=yes` was invalid due to the `&&`, but this is fine because this just means that there's an empty query parameter is ignored

### DIFF
--- a/modules/ivts/galasa-ivts-parent/dev.galasa.ivts/dev.galasa.ivts.http/src/main/java/dev/galasa/ivts/http/HttpManagerIVT.java
+++ b/modules/ivts/galasa-ivts-parent/dev.galasa.ivts/dev.galasa.ivts.http/src/main/java/dev/galasa/ivts/http/HttpManagerIVT.java
@@ -227,27 +227,17 @@ public class HttpManagerIVT {
     }
     
     @Test
-    public void buildURITest() {
-    	HttpClientException expected = null;
-    	
-    	try {
-    		// dummy request
-    		client.getText("http://httpbin.org/anything?should==fail");
-    	} catch (HttpClientException e) {
-    		logger.info("Caught expected exception: " + e.getMessage());
-    		expected = e;
-    	}
-    	assertThat(expected).isNotNull();
-    	
-    	expected = null;
-    	try {
-    		// dummy request
-    		client.getText("http://httpbin.org/anything?thisparam=ok&&oops=yes");
-    	} catch (HttpClientException e) {
-    		logger.info("Caught expected exception: " + e.getMessage());
-    		expected = e;
-    	}
-    	assertThat(expected).isNotNull();
+    public void buildURITest() throws Exception {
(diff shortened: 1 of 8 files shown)


modules/ivts/galasa-ivts-parent/dev.galasa.ivts/dev.galasa.ivts.http/src/main/java/dev/galasa/ivts/http/HttpManagerIVT.java
```diff
@@ -227,27 +227,17 @@ public void downloadFileTest() throws Exception {
...
-    		client.getText("http://httpbin.org/anything?thisparam=ok&&oops=yes");
-    	} catch (HttpClientException e) {
-    		logger.info("Caught expected exception: " + e.getMessage());
-    		expected = e;
-    	}
-    	assertThat(expected).isNotNull();
+    public void buildURITest() throws Exception {
+    	// Query parameter with double equals: "should==fail"
+    	// This is valid - parsed as parameter "should" with value "=fail"
+    	HttpClientResponse<String> response1 = client.getText("http://httpbin.org/anything?should==fail");
```
jadecarino: This is super nitpicky as its dummy test data but can we please change this to `should==pass` now that this test should pass? If this were to fail (incorrectly) and then it came up in the debug log and I saw the `should==fail`, it might be confusing as to what the behaviour should be.

jadecarino: Same in the other HttpManagerIVT file

jadecarino (approved):
Just one small change but otherwise looks good!

modules/ivts/galasa-ivts-parent/dev.galasa.ivts/dev.galasa.ivts.http/src/main/java/dev/galasa/ivts/http/HttpManagerIVT.java
eamansour: Great spot - addressed in https://github.com/galasa-dev/galasa/pull/578/commits/0f7a9b6fe4953d07231b156829beff68f96a713c

PR #579, opened 2026-04-21
Iss2573 - Support KeyStore type secrets in `galasactl secrets set` 

jadecarino:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2573

- Updates to `galasactl secrets set` to be able to create/update KeyStore secrets.
   - New flags to provide the KeyStore data have been added to the command: either from a .p12 or .jks file or a base64 encoded version of this file (`--keystore-file`, `--base64-keystore-encoded`)
   - The keystore password is provided with the existing `--password` or `--base64-password` flags, and optionally the keystore type with `--keystore-type` (defaults to PKCS12 if not provided)
   - The values for these keystore related flags are put into a struct so only one new parameter needs to get passed through the code and validation of the fields can happen on the struct instead of in functions throughout the code.
- `galasactl secrets get [--name MY_KEYSTORE]` already worked with no code changes as the secrets formatter simply unmarshals the GalasaSecret data into either summary or yaml format using the payload from the API, but unit tests have been added.
- To remove CodeQL error, I have removed the explicit logging of the `err` in secretsSet.go - if there is an error, it is still returned.

## Changes


jadecarino:
@eamansour Hi Eamonn, as mentioned on Slack, the whitespace for the file secretsSet.go was altered so functions that have changes that actually need reviewing are: SetSecret, createSecretRequestKeystore, createSecretRequestKeystorePassword, createSecretRequest, validateFlagCombination, validateKeystorePassword and validateKeystoreType. The same is with secretsSet_test.go - new unit tests are from line 1417 onwards.

modules/cli/docs/generated/errors-list.md
```diff
@@ -290,6 +290,10 @@ The `galasactl` tool can generate the following errors:
...
+- GAL1291E: Invalid keystore type provided. The keystore type must be either 'JKS' or 'PKCS12'. Use the --help flag for more information, or refer to the documentation at https://galasa.dev/docs/reference
+- GAL1292E: Missing required keystore password. When creating a keystore secret using --keystore-file, --keystore-encoded, or --base64-keystore-encoded, you must also provide --keystore-password or --base64-keystore-password. Use the --help flag for more information, or refer to the documentation at https://galasa.dev/docs/reference
+- GAL1293E: Invalid keystore flag combination provided. --keystore-file cannot be provided with --keystore-encoded or --base64-keystore-encoded, and --keystore-password cannot be provided with --base64-keystore-password. Use the --help flag for more information, or refer to the documentation at https://galasa.dev/docs/reference
+- GAL1294E: Empty keystore file provided. The keystore file '{}' contains 0 bytes. Please provide a valid keystore file. Use the --help flag for more information, or refer to the documentation at https://galasa.dev/docs/reference
```
eamansour: If the location of the CLI reference ever changes, then we'd have to update these messages as well

Would it be better to remove the links to the galasa.dev site in these error messages?

modules/cli/pkg/cmd/secretsSet.go
```diff
@@ -26,6 +26,12 @@ type SecretsSetCmdValues struct {
     password string
     token string
 	description string
+	keystoreEncoded string
+	keystoreFile string
+	keystorePassword string
+	base64KeystoreEncoded string
```
eamansour: If keystores are already base64 encoded, then I think we don't need two separate flags for supplying the encoded value.

Maybe we could support `--base64-keystore-encoded` and remove `--keystore-encoded`?

modules/cli/pkg/cmd/secretsSet.go
```diff
@@ -99,12 +105,19 @@ func (cmd *SecretsSetCommand) createCobraCmd(
     usernameFlag := "username"
     passwordFlag := "password"
     tokenFlag := "token"
-
+    keystoreEncodedFlag := "keystore-encoded"
+    keystoreFileFlag := "keystore-file"
+    keystorePasswordFlag := "keystore-password"
```
eamansour: Are we able to reuse the `--password` flag for keystores? Would that make sense from a user's point of view?

So someone could do:

```
galasactl secrets set --name x --keystore-file y --password z
```

Just thinking ahead, if we had other secret types that required passwords, then we'd end up with a lot of `<secret-type>-password` flags.

modules/cli/pkg/secrets/secretsSet.go
```diff
@@ -7,217 +7,347 @@
...
+	token string,
+	base64Username string,
+	base64Password string,
+	base64Token string,
+	keystoreEncoded string,
+	keystoreFile string,
+	keystorePassword string,
+	base64KeystoreEncoded string,
+	base64KeystorePassword string,
+	keystoreType string,
```
eamansour: To avoid this parameter list from becoming too long, we could perhaps place the keystore-related parameters into their own struct and then pass that struct into this function?

modules/cli/pkg/secrets/secretsSet.go
```diff
@@ -7,217 +7,347 @@
...
+}
+
+func validateKeystoreType(keystoreType string) error {
+	var err error
+	validTypes := []string{"JKS", "PKCS12"}
+	if !slices.Contains(validTypes, strings.ToUpper(keystoreType)) {
+		err = galasaErrors.NewGalasaError(galasaErrors.GALASA_ERROR_INVALID_KEYSTORE_TYPE)
+	}
+	return err
+}
```
eamansour: If you decide to put the keystore parameters into their own struct, these validation functions could also go inside that struct so that they can be called as part of a `validate()` function like:

```
keystoreStruct.validate()
```

eamansour (commented):
Looks good overall, just a few minor suggestions

modules/cli/docs/generated/errors-list.md
jadecarino: Good point thanks Eamonn, updated in https://github.com/galasa-dev/galasa/pull/579/commits/9d33ff11bff64cff20dbebe26ce95db3ec411606

modules/cli/pkg/cmd/secretsSet.go
jadecarino: Yes I was wondering about this! I have removed --keystore-encoded in favour of --base-64-encoded in https://github.com/galasa-dev/galasa/pull/579/commits/07f1b3827e56634d003df7746c4d78c1df76ee49

modules/cli/pkg/cmd/secretsSet.go
jadecarino: Great suggestion, I've updated this in https://github.com/galasa-dev/galasa/pull/579/commits/07f1b3827e56634d003df7746c4d78c1df76ee49

modules/cli/pkg/secrets/secretsSet.go
jadecarino: Thanks, I've updated this in https://github.com/galasa-dev/galasa/pull/579/commits/07f1b3827e56634d003df7746c4d78c1df76ee49

modules/cli/pkg/secrets/secretsSet.go
jadecarino: Thanks, I've updated this in https://github.com/galasa-dev/galasa/commit/07f1b3827e56634d003df7746c4d78c1df76ee49

eamansour (approved):
Looks good! Thanks for making those changes!

PR #580, opened 2026-04-23
Replace mentions of galasa-service1 with galasa-service-main

eamansour:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/1531.

## Changes
- [x] Replaced mentions of galasa-service1 with galasa-service-main


jadecarino (approved):
Great stuff! This will make things much easier in future

PR #588, opened 2026-04-27
galasa.dev documentation improvements

jadecarino:
## Why?

Improvements made to the external docs made based on comments on our call @eamansour 

## Changes
Offline docs:
- [x] Relocation of the docs for using Galasa offline/the zipped distribution into its own section that follows the SimBank section, to reduce the amount of information first-time users are being presented with in the Getting Started section that may not be relevant.
- [x] A link to this section has been provided at the end of the Getting Started page though to signpost to it for users who do work in a restrictive network environment and do want to use Galasa offline.
- [x] Due to this change, I've removed places in the Getting Started pages that refer to using Galasa offline to reduce noise.
- [x] Deleted the prerequisites page for the zipped distribution as its exactly the same for online.

Formatting:
- [x] Mkdocs tabs have been added to allow users to switch back and forth for commands for Linux vs MacOS vs Windows instead of having bold headings. This reduces the length of the rendered pages greatly.

Getting Started:
- [x] In the page for downloading the CLI, links to the latest released version have been provided in the table so users can simply click the binary they need. **Note**, this has not been included in the 'offline' equivalent as the binaries are included in the zip.

SimBank:
- [x] Slight rewrite in the page to introduce SimBank to make it clearer what it is and what it is intended for. We can continue this further in a future PR.


docs/content/docs/cli-command-reference/index.md
```diff
@@ -2,20 +2,36 @@
...
+- Installing the required software prerequisites
+- Downloading and installing the Galasa CLI
+- Initializing your local Galasa environment
+- Creating and running Galasa test projects
+- Viewing test results
 
-The following topics in this section take you through the software requirements for each install option, and explain how to get started with the galasactl tool. You can then find out how to initialise your local environment so that you can start creating, building and running Galasa tests, and viewing the test run output. 
+## Getting started
 
+Start by installing the necessary software prerequisites, then download and install the Galasa CLI from the [Galasa repository](https://github.com/galasa-dev/galasa/releases){target="_blank"} on GitHub. Once installed, you can initialize your local environment and begin creating, building, and running tests.
```
eamansour: Not sure if this paragraph is needed since the steps below list everything users need to do.

Also, this mentions downloading the CLI from GitHub but later we go through how to download it via brew, so users would end up having to delete the downloaded CLI if they want to use brew.

docs/content/docs/cli-command-reference/installing-cli-tool.md
```diff
@@ -1,82 +1,111 @@
...
+       ```
+    
+    2. Install the latest version:
+       ```bash
+       brew install --no-quarantine galasactl
+       ```
+       
+       Or install a specific version (e.g., 0.48.0):
+       ```bash
+       brew install --no-quarantine galasactl@0.48.0
```
eamansour: `--no-quarantine` is deprecated, we should be able to use brew without the flag but will need to run the `xattr` command for macOS

eamansour (approved):
A couple of minor suggestions but overall looks good, happy to approve!

docs/content/docs/cli-command-reference/index.md
jadecarino: Great spot thanks, removed in https://github.com/galasa-dev/galasa/pull/588/commits/3efa455333f9648f8895a1845af33be15b9f4400

docs/content/docs/cli-command-reference/installing-cli-tool.md
jadecarino: Yes great point thank you, removed the flag and added the xattr command in https://github.com/galasa-dev/galasa/pull/588/commits/8454072730533a93f24222b6798a98d7c7892dc1

PR #590, opened 2026-05-12
Start simplifying the 'Getting started' docs and use collapsible blocks for optional content

eamansour:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2581

## Changes
- [x] Started simplifying the content on the following pages:
  - [x] "Initialising your local environment"
  - [x] "Installing the Galasa CLI"
  - [x] "Running a test locally" (note: we still need to separate the Gherkin-related content from this page out into a separate topic as we don't cover writing Gherkin tests anywhere)
- [x] Moved "Setting the Galasa home folder (optional)" into a collapsible block
- [x] Moved "Overriding the path to the default local Maven repository" into a collapsible block (since this is optional and users can just use the default `.m2` repo)
- [x] Removed the "Setting the Galasa bootstrap" section from the "Initialising your local environment" page since this focuses on setting `GALASA_BOOTSTRAP` to an ecosystem's bootstrap endpoint (which is not something users would do when first getting started with Galasa)


docs/content/docs/cli-command-reference/initialising-home-folder.md
```diff
@@ -50,102 +53,85 @@ ${HOME}/.galasa
...
-
-=== "Windows (Powershell)"
-
-    ```powershell
-    set GALASA_BOOTSTRAP=http://galasa-bootstrap.url.com/bootstrap
-    ```
-
-
+| **bootstrap.properties**  | Contains the connection details for a Galasa Ecosystem (a shared test environment). This file is empty when running tests locally without shared configuration. |
+| **cps.properties** | Stores configuration settings for your tests, such as endpoints, ports, and timeouts. These settings let you run the same test against different environments without changing your test code. Use the `@TestProperty` annotation in your tests to read values from this file. |
```
jadecarino: I wonder whether we should say `for your remote test systems` here instead of `for your tests` so it doesn't seem that we mean these will change test execution (like a timeout for the tests?)

docs/content/docs/cli-command-reference/setting-up-galasa-project.md
```diff
@@ -4,64 +4,266 @@ title: "Creating a Galasa project"
...
+    repositories {
+        mavenLocal()
+        mavenCentral()
+    }
+
+    group = 'dev.galasa.example.banking'
+    version = '0.0.1-SNAPSHOT'
+
+    dependencies {
+        implementation platform('dev.galasa:galasa-bom:0.48.0')
```
jadecarino, line 153: Could we add something to the `set-version.sh` to keep the versions in the examples up to date please?

jadecarino (commented):
Two minor suggestions but otherwise I've had a look at the site with changes locally and looks great!

docs/content/docs/cli-command-reference/initialising-home-folder.md
eamansour: Good spot! Addressed in https://github.com/galasa-dev/galasa/pull/590/commits/cc927c123312c33a30960955e3a8f1fe7dbc2731

docs/content/docs/cli-command-reference/setting-up-galasa-project.md
eamansour: Updated the set-version script in https://github.com/galasa-dev/galasa/pull/590/commits/cc927c123312c33a30960955e3a8f1fe7dbc2731

jadecarino (approved):

PR #592, opened 2026-05-13
Simplify the index page that is presented when navigating to the docs

eamansour:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2581.

## Changes
- [x] Rewrote the index page that presents users with quick start information to Galasa and an overview of the various Galasa components
- [x] Moved the architecture diagram caption to be below the image

### DIFF
--- a/docs/content/docs/index.md
+++ b/docs/content/docs/index.md
@@ -2,46 +2,153 @@
 title: Galasa Documentation
 icon: material/book-open-variant
 ---
- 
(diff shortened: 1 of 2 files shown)


jadecarino (approved):
Changes look good, although when I ran the site locally, the search bar in the header had disappeared. Not sure if this happens everytime you build and run the site locally if so ignore!

PR #593, opened 2026-05-13
HTTP Manager: TLS + postForm + timeout + clearcookie + redirectstrategy

Mark-J-Lawrence:
## Why?

This PR delivers a number of changes which have been required whilst migrating our tests over from using the v4 HTTP client, and our own internal v5 HTTP client, over to the new public httpclient5 HTTP manager delivered in 0.47.0.

The changes are:

1. New text content types available to support return types by springboot 4 applications
2. The ability to set the c**onnection request** + **response** timeout manually via annotation metadata
3. The ability to set the **socket** and **connection** timeout via helper functions
4. A new `postForm()` handler which makes working with encoded forms a lot easier (simply provide a url and a key-value map for the form)
5. Latest TLS versions supported for the connection
6. A helper to clear cookies, which can be handy when testing authentication scenarios.
7. Better redirect support. Firstly, to enable it, secondly, a redirect strategy to handle when a redirect after authentication may go back to `/`, however, the trailing `/` is lost and it simply goes back to the baseUrl, causing failures. 

## Changes
- [ ] Unit tests (if applicable)

### DIFF
--- a/modules/managers/galasa-managers-parent/galasa-managers-comms-parent/dev.galasa.http.manager/src/main/java/dev/galasa/http/IHttpClient.java
+++ b/modules/managers/galasa-managers-parent/galasa-managers-comms-parent/dev.galasa.http.manager/src/main/java/dev/galasa/http/IHttpClient.java
@@ -208,6 +208,28 @@ public interface IHttpClient {
     HttpClientResponse<String> postText(String url, String text) throws HttpClientException;
 
     /**
+     * Issue an HTTP POST to the provided URL, sending form-encoded data
+     * and receiving a {@link String} in the response.
+     *
+     * <p>
+     * This is a convenience method for posting form data (e.g., login credentials).
+     * The data is encoded as application/x-www-form-urlencoded, which is the standard
+     * format for HTML form submissions.
+     * </p>
+     *
+     * <p>
+     * This method is commonly used for form-based authentication where credentials
+     * are posted as form fields (e.g., j_username, j_password for Java EE security).
+     * </p>
+     *
+     * @param url - URL path
+     * @param fields - Form fields as key-value pairs
+     * @return - {@link HttpClientResponse} with a {@link String} content type
+     * @throws HttpClientException if the request fails
+     */
+    HttpClientResponse<String> postForm(String url, Map<String, String> fields) throws HttpClientException;
+
+    /**
      * Issue an HTTP PUT to the provided URL, sending the provided {@link String}
      * and receiving a {@link String} in the response.
      * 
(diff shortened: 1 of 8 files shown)


modules/managers/galasa-managers-parent/galasa-managers-comms-parent/dev.galasa.http.manager/src/main/java/dev/galasa/http/IHttpClient.java
```diff
@@ -508,5 +548,18 @@ IHttpClient setupClientAuth(KeyStore keyStore, String password)
...
+     * Clear the cookie store
+     */
+    void clearCookies();
+    
+    /**
+     * Set TLS versions required.
+     *
+     * @param tlsVersions
+     *            TLS versions
+     */
```
eamansour, line 562: Are there any requirements around what format the TLS version strings should be given in? ie. will they always follow the form `TLSvX` (where `X` is the version number)?

Would it be worth creating our own TLSVersion enum that users can pick from?

modules/managers/galasa-managers-parent/galasa-managers-comms-parent/dev.galasa.http.manager/src/main/java/dev/galasa/http/internal/HttpClientImpl.java
```diff
@@ -967,4 +1048,48 @@ public void close() {
...
+    @Override
+    public void clearCookies() {
+        this.cookieStore.clear();
+    }
+
+    @Override
+    public void setTlsVersions(String[] tlsVersions)
+    {
+        this.tlsVersions = Arrays.copyOf(tlsVersions, tlsVersions.length);
+    }
```
eamansour, line 1065: Incredibly minor - could we keep the brackets consistent with the rest of the code (ie. opening brackets on the same line as function/condition declarations)?

Same applies throughout this PR

(we still need to add styling checks to enforce this better!)

eamansour (commented):
Thanks for contributing these changes, Mark! They look great overall, just a couple of very minor comments - I'll be happy to approve once they're sorted out

modules/managers/galasa-managers-parent/galasa-managers-comms-parent/dev.galasa.http.manager/src/main/java/dev/galasa/http/IHttpClient.java
Mark-J-Lawrence: No probs, added new dev.galasa.http.TLS class and updated throughout

modules/managers/galasa-managers-parent/galasa-managers-comms-parent/dev.galasa.http.manager/src/main/java/dev/galasa/http/internal/HttpClientImpl.java
Mark-J-Lawrence: No probs, I've update everywhere I've spotted this, let me know if I've missed any!

eamansour (approved):

PR #607, opened 2026-05-22
Iss2571 - Improve CouchDB Users and Tokens Views Performance

### LINKED ISSUE #2571: Performance: Avoid emitting entire document from CouchDB view queries for users/tokens
## Story
As a Galasa service user, I want the users/tokens views from CouchDB to only emit the whole document if its required, so that view queries do not use up unnecessary disk space and cause performance issues.

## Background
This story comes from a suggestion in a review [comment](https://github.com/galasa-dev/galasa/pull/562#discussion_r3029299011) in [PR](https://github.com/galasa-dev/galasa/pull/562). The existing views had been emitting the whole document as does the new view introduced in the PR. All views should be updated to not emit the whole view and if needed, pass in the `include_docs` query parameter in the request.

## Tasks
- [ ] Update all three views in CouchdbAuthStoreValidator to not emit the whole document
- [ ] Update the method that makes the request to the Auth store to accept a boolean for includeDocs
- [ ] If boolean is true, add the `include_docs` query parameter in the request
- [ ] Test on Minikube that all requests involving these views work as expected

jadecarino:
## Why?

Fixes https://github.com/galasa-dev/projectmanagement/issues/2571

The User and Token Couchdb views will not emit the whole document unless the `include_docs=true` parameter is added to the request to avoid using excess disk space which could cause performance issues.

## Changes
- [x] Users and Tokens Couchdb Views updated
- [x] `getAllDocsByLoginId` takes a boolean parameter of whether to add the `include_docs=true` parameters to the request
- [x] Changes tested on Minikube with `galasactl users` and `auth tokens` commands
- [x] Verified that there are not duplicate views (old and new) in `_design/docs` for galasa_users and galasa_tokens in my Minikube system

### DIFF
--- a/modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
+++ b/modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
@@ -254,7 +254,13 @@ public class CouchdbAuthStore extends CouchdbStore implements IAuthStore {
 
         try {
             // Get all of the documents in the tokens database with include_docs=true
-            tokenDocuments = getAllDocsByLoginId(TOKENS_DATABASE_NAME, lowercaseLoginId, TOKENS_DB_VIEW_NAME, true);
+            ViewResponse viewResponse = getDocumentsFromDatabaseViewByKey(
+                TOKENS_DATABASE_NAME,
+                TOKENS_DB_VIEW_NAME,
+                lowercaseLoginId,
+                true
+            );
+            tokenDocuments = viewResponse.rows;
 
             // Build up a list of all the tokens using the documents from the view response
             for (ViewRow row : tokenDocuments) {
@@ -403,7 +409,13 @@ public class CouchdbAuthStore extends CouchdbStore implements IAuthStore {
             String lowerCaseLoginId = loginId.toLowerCase();
             
             // Fetch documents matching the lowercase loginId using the lowercase view with include_docs=true
-            userDocument = getAllDocsByLoginId(USERS_DATABASE_NAME, lowerCaseLoginId, USERS_DB_VIEW_NAME, true);
+            ViewResponse viewResponse = getDocumentsFromDatabaseViewByKey(
+                USERS_DATABASE_NAME,
+                USERS_DB_VIEW_NAME,
+                lowerCaseLoginId,
+                true
+            );
+            userDocument = viewResponse.rows;
 
             // Since loginIds are unique (case-insensitive), there should be only one document.
             if (userDocument != null && !userDocument.isEmpty()) {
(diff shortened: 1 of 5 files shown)


modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -446,30 +448,43 @@ public IUser updateUser(IUser user) throws AuthStoreException {
...
-     * @param loginId the loginId of the user to retrieve the doucemnts of
+     * @param loginId the loginId of the user to retrieve the documents of
+     * @param viewName the name of the view to query
+     * @param includeDocs whether to include the full document in the response using include_docs=true
      * @return a list of rows corresponding to documents within the database
      * @throws CouchdbException if there was a problem accessing the
      *                          CouchDB store or its response
      */
-    protected List<ViewRow> getAllDocsByLoginId(String dbName, String loginId, String viewName)
+    protected List<ViewRow> getAllDocsByLoginId(String dbName, String loginId, String viewName, boolean includeDocs)
```
eamansour: This method now duplicates logic that's already in the `getDocumentsFromDatabaseViewByKey` method (available in the CouchdbStore parent class) - can we use that instead?

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
```diff
@@ -446,30 +448,43 @@ public IUser updateUser(IUser user) throws AuthStoreException {
...
-        String responseEntity = sendHttpRequest(getDocs, HttpStatus.SC_OK);
-
-        ViewResponse docByLoginId = gson.fromJson(responseEntity, ViewResponse.class);
-        List<ViewRow> viewRows = docByLoginId.rows;
-
-        if (viewRows == null) {
-            String errorMessage = ERROR_FAILED_TO_GET_DOCUMENTS_FROM_DATABASE.getMessage(dbName);
-            throw new CouchdbException(errorMessage);
+        String viewRequestUriStr = storeUri + "/" + dbName + "/_design/docs/_view/" + viewName;
+        // String encodedLoginId = URLEncoder.encode("\"" + loginId + "\"", StandardCharsets.UTF_8);
```
eamansour: Can this commented-out code be removed if it's not needed?

eamansour (commented):
Looks good, just one main suggestion to reduce duplicate code

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
jadecarino: Great point! I've removed getAllDocsByLoginId and the auth store code now uses getDocumentsFromDatabaseViewByKey as well.

modules/extensions/galasa-extensions-parent/dev.galasa.auth.couchdb/src/main/java/dev/galasa/auth/couchdb/internal/CouchdbAuthStore.java
jadecarino: Good catch - method deleted so this has been removed.

eamansour (approved):

PR #627, opened 2026-06-03
Add Windows Credential Manager store implementation

eamansour:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2592

## Changes
- [x] Added an implementation of the credentials store that uses the Windows Credential Manager, allowing for read-only access to credentials in tests
  - Note: this implementation uses JNA to access the Windows APIs rather than spawning a new shell process to run commands, which can be expensive and slow

### DIFF
--- a/modules/extensions/galasa-extensions-parent/dev.galasa.creds.os/src/test/java/dev/galasa/creds/os/internal/windows/WindowsCredentialManagerTest.java
+++ b/modules/extensions/galasa-extensions-parent/dev.galasa.creds.os/src/test/java/dev/galasa/creds/os/internal/windows/WindowsCredentialManagerTest.java
@@ -64,40 +64,6 @@ public class WindowsCredentialManagerTest {
     }
 
     @Test
-    public void testReadCredentialWithUsernameOnly() throws OsCredentialsException {
-        // Given
-        String targetName = "galasa.credentials.USERONLY";
-        String username = "testuser";
-        mockJNA.addCredential(targetName, username, "");
-
-        // When
-        CredentialItem result = credentialManager.readCredential(targetName);
-
-        // Then
-        assertThat(result).isNotNull();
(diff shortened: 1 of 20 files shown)


modules/extensions/galasa-extensions-parent/dev.galasa.creds.os/src/test/java/dev/galasa/creds/os/internal/windows/WindowsCredentialManagerTest.java
```diff
@@ -0,0 +1,373 @@
...
+        // Then
+        assertThat(result).isNotNull();
+        assertThat(result.getUsername()).isEmpty();
+        assertThat(result.getPassword()).isEqualTo("password");
+        assertThat(mockJNA.isMemoryFreed()).isTrue();
+    }
+
+    @Test
+    public void testReadCredentialWithEmptyPassword() throws OsCredentialsException {
+        // Given
```
jadecarino, line 236: These seem to be duplicates of the test cases `testReadCredentialWithUsernameOnly` and `testReadCredentialWithPasswordOnly`, should we remove for conciseness?

jadecarino (approved):
One minor suggestion but otherwise looks great!

modules/extensions/galasa-extensions-parent/dev.galasa.creds.os/src/test/java/dev/galasa/creds/os/internal/windows/WindowsCredentialManagerTest.java
eamansour: Agreed - removed in https://github.com/galasa-dev/galasa/pull/627/commits/78dab6aed174e8718dc9f55fbabaf5d18b651829

jadecarino (approved):

PR #642, opened 2026-06-19
Allow multiple --class flags in runs submit local commands

eamansour:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2610

## Changes
- [x] The framework now creates `felix-cache` directories under a common `cache` folder within the Galasa home directory
- [x] The CLI no longer "loses" test runs when multiple `--class` flags are given
- [x] `runs submit local` now sets a default throttle of 1 to avoid JVMs from competing with each other over resources and DSS changes


jadecarino (approved):
Approved - thanks for fixing the issue with the run name not being able to be identified also!

PR #643, opened 2026-06-22
Pre-compute priority scores before sorting to avoid timeService calls from breaking comparator contract

eamansour:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2611

As part of the priority scheduling code, every run comparison call goes through:

```
getQueuedRunTotalPriorityPoints() -> getPriorityPointsFromQueuedTime() -> timeService.now()
```

This likely means that the clock advances between comparisons during a sort, and in turn, means the score for the same run object can return a different value in successive calls. This violates Java's comparator contract (transitivity/antisymmetry), hence the `IllegalArgumentException`.

This PR solves this problem by pre-computing the priority score values for the queued test runs, so that the priority scores cannot change during sorting.

## Changes
- [x] Updated the priority scheduling code to pre-compute queued run priority points and pass the pre-computed values in the sort, rather than computing the priority points during sorting


jadecarino (approved):
Very nice!

PR #653, opened 2026-06-30
Add JMeterManagerIVT to ivts module

jadecarino:
## Why?

This PR adds the JMeterManagerIVT to the ivts module so that it can be ran remotely in a Galasa service from the ivts test stream. I'll open two additional PRs:
- One with a GH Actions workflow to install a JMeter binary and run this on service-main in `LOCAL` mode
- One to run this on prod1 in `DOCKER` mode connecting to our protected docker engine (must be prod1 as its internal infrastructure)

**Note:** @eamansour  It feels to me we should remove all IVT bundles (e.g., dev.galasa.jmeter.manager.ivt) from the managers module as they need to be replicated in the ivts module to be ran remotely. Them being in the managers doesn't add anything and just doubles maintenance. Is there a reason we might want to keep them in managers?

## Changes
- [x] Add dev.galasa.ivts.jmeter to the ivts module with the JMeterManagerIVT

### DIFF
--- a/modules/ivts/galasa-ivts-parent/dev.galasa.ivts/dev.galasa.ivts.jmeter/bnd.bnd
+++ b/modules/ivts/galasa-ivts-parent/dev.galasa.ivts/dev.galasa.ivts.jmeter/bnd.bnd
@@ -1,0 +1,5 @@
+-snapshot: ${tstamp}
+Bundle-Name: Galasa JMeter Manager IVT
+Export-Package: dev.galasa.jmeter.manager.ivt
+Import-Package: !javax.validation.constraints, \
+                *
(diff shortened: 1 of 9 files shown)


eamansour (approved):
Looks good, thanks for adding this IVT here. It definitely makes sense to run this IVT more regularly in a workflow.

> Note: @eamansour It feels to me we should remove all IVT bundles (e.g., dev.galasa.jmeter.manager.ivt) from the managers module as they need to be replicated in the ivts module to be ran remotely. Them being in the managers doesn't add anything and just doubles maintenance. Is there a reason we might want to keep them in managers?

I don't see a reason to keep IVTs in the managers module, so I'm more than happy for the ones that have been added to the ivts module to be removed.

PR #654, opened 2026-06-30
Separate jetcd-core jars from Galasa cps.etcd bundle to its own wrapped bundle

eamansour:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2608

This PR creates a new wrapping bundle for the non-OSGi jetcd-core library, separating the library out from the CPS etcd extension.

## Changes
- [x] Added dev.galasa.wrapping.jetcd-core wrapping bundle
- [x] Removed explicit `-includeresource` jars from the CPS etcd extension's bnd.bnd file
- [x] Removed explicit transitive dependencies listed in the CPS etcd extension's build.gradle file
- [x] Removed transitive dependencies that were specific to the jetcd library from the gradle platform


jadecarino (approved):
Much better!

PR #666, opened 2026-07-09
Iss2616 - KeyStore credentials can now have an empty-string password

jadecarino:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2616

- Added support for an empty-string `keystorePassword` (`""`) on both create and update paths, to indicate a keystore with no integrity-check password — whitespace-only values are still rejected
- Added server-side validation (GAL5462) rejecting updates to an existing KeyStore credential where only one of `keystore` or `keystorePassword` is provided — both must be supplied together. Required to validate the KeyStore data and password together upon creation/update and avoids ambiguity with the CLI flag `--password` carrying both normal and KeyStore passwords.
- `galasactl secrets set`: `--password ""` is now accepted alongside `--keystore-file`/`--base64-keystore-encoded` and is correctly serialised into the request body to intentionally set an empty password 
- Updated GAL5453 and GAL1292E error messages to document that `""` is a valid `keystorePassword` value
- Updated README.md to document the empty-password behaviour and the constraint that `--keystore-file`/`--base64-keystore-encoded` and `--password` must be supplied together on update
- Added release note to v1.0.0

## Changes


modules/framework/galasa-parent/dev.galasa.framework.api.common/src/main/java/dev/galasa/framework/api/common/ServletErrorMessage.java
```diff
@@ -162,7 +162,8 @@ public enum ServletErrorMessage {
...
     GAL5451_MUTUALLY_EXCLUSIVE_FIELDS_PROVIDED        (5451, "E: Invalid secret payload provided. The ''{0}'' field cannot be used with KeyStore credentials. KeyStore credentials only accept ''keystore'', ''keystoreType'', and ''keystorePassword'' fields. Check your request payload and try again."),
     GAL5452_INVALID_KEYSTORE_BASE64_ENCODING          (5452, "E: Invalid keystore value provided. The keystore value must be valid base64-encoded data. Check your request payload and try again."),
-    GAL5453_MISSING_KEYSTORE_PASSWORD_FIELD           (5453, "E: Invalid secret payload provided. The ''keystorePassword'' field is missing. Check your request payload and try again."),
+    GAL5453_MISSING_KEYSTORE_PASSWORD_FIELD           (5453, "E: Invalid secret payload provided. The ''keystorePassword'' field is missing. An empty string \"\" is a valid value to indicate no password. Check your request payload and try again."),
+    GAL5462_KEYSTORE_AND_PASSWORD_MUST_BE_PROVIDED_TOGETHER (5462, "E: Invalid secret payload provided. When updating a KeyStore credential, the ''keystore'' and ''keystorePassword'' fields must be provided together. Check your request payload and try again."),
```
eamansour, line 166: If someone only wants to update a keystore type (e.g. from `PKS12` to `JKS`) by only supplying the secret name and the keystore type, would that now fail as the payload expects the `keystore` and `password`?

We might need a new unit test to check that keystore type can be provided on its own if we still want to support partial updates for keystores (assuming there isn't one already!).

Also, does the openapi.yaml mention this requirement on `keystore` and `keystorePassword` being provided together?

eamansour (commented):
Just one comment, but overall looks good!

modules/framework/galasa-parent/dev.galasa.framework.api.common/src/main/java/dev/galasa/framework/api/common/ServletErrorMessage.java
jadecarino: Thanks Eamonn, the keystoreType can be supplied on its own, and I have added a unit test for that specifically and more description about this in https://github.com/galasa-dev/galasa/pull/666/commits/ccd5984fbe7665dd2f6e3a8476fcf8e806ba534c

eamansour (approved):

PR #670, opened 2026-07-17
Iss2616 - Binary secret data can be stored in the Galasa credentials store

jadecarino:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2616

Adds `Opaque` credential type to the Galasa Credentials Store allowing users to store and retrieve opaque binary data such as licence JAR files or certificates for dynamic use during test runs.

Two new `galasactl secrets set` flags:`--secret-file` (reads a file from disk and base64-encodes it) and `--base64-secret` (accepts an already-encoded string) are mutually exclusive with each other and with all other credential flags.

## Changes


modules/cli/pkg/cmd/secretsSet.go
```diff
@@ -127,18 +133,27 @@ func (cmd *SecretsSetCommand) createCobraCmd(
...
+    secretsSetCobraCmd.Flags().StringVar(&cmd.values.binaryValues.Base64BinaryEncoded, base64BinaryEncodedFlag, "", "a base64-encoded binary value to set into a secret")
+
     // A non-encoded credential cannot be provided alongside an encoded credential
     secretsSetCobraCmd.MarkFlagsMutuallyExclusive(usernameFlag, base64UsernameFlag)
 
     // Keystore file and keystore encoded are mutually exclusive (can only provide one way to specify keystore)
     secretsSetCobraCmd.MarkFlagsMutuallyExclusive(keystoreFileFlag, base64KeystoreEncodedFlag)
 
+    // Binary file and binary encoded are mutually exclusive (can only provide one way to specify binary data)
+    secretsSetCobraCmd.MarkFlagsMutuallyExclusive(binaryFileFlag, base64BinaryEncodedFlag)
```
eamansour: Minor - we can remove this `MarkFlagsMutuallyExclusive` line since we're setting it again further down in:

```go
    // Binary cannot be provided alongside any other credential field
    secretsSetCobraCmd.MarkFlagsMutuallyExclusive(binaryFileFlag, base64BinaryEncodedFlag, usernameFlag, base64UsernameFlag, passwordFlag, base64PasswordFlag, tokenFlag, base64TokenFlag, keystoreFileFlag, base64KeystoreEncodedFlag)
```

modules/cli/pkg/secrets/secretsSet.go
```diff
@@ -73,13 +75,23 @@ func SetSecret(
...
 
+				// Handle binary data - either from file or encoded string
+				if err == nil {
+					if binaryValues.BinaryFile != "" || binaryValues.Base64BinaryEncoded != "" {
+						requestBinary, err = createSecretRequestBinary(binaryValues.BinaryFile, binaryValues.Base64BinaryEncoded, fileSystem)
+					}
+				}
+				// if err == nil && (binaryValues.BinaryFile != "" || binaryValues.Base64BinaryEncoded != "") {
+				// 	requestBinary, err = createSecretRequestBinary(binaryValues.BinaryFile, binaryValues.Base64BinaryEncoded, fileSystem)
+				// }
```
eamansour: Can we remove this commented-out code?

modules/cli/pkg/cmd/secretsSet.go
```diff
@@ -127,18 +133,27 @@ func (cmd *SecretsSetCommand) createCobraCmd(
     secretsSetCobraCmd.Flags().StringVar(&cmd.values.keystoreValues.KeystoreType, keystoreTypeFlag, "", "the type of keystore (PKCS12 or JKS). Defaults to PKCS12 if not specified")
     secretsSetCobraCmd.Flags().StringVar(&cmd.values.keystoreValues.Base64KeystoreEncoded, base64KeystoreEncodedFlag, "", "a base64-encoded keystore to set into a secret")
 
+    secretsSetCobraCmd.Flags().StringVar(&cmd.values.binaryValues.BinaryFile, binaryFileFlag, "", "the path to a binary file (e.g. a .jar or certificate) to set into a secret")
+    secretsSetCobraCmd.Flags().StringVar(&cmd.values.binaryValues.Base64BinaryEncoded, base64BinaryEncodedFlag, "", "a base64-encoded binary value to set into a secret")
```
eamansour: Are there any limits as to how large the binary data can be? Would etcd be able to store large files?

docs/content/releases/posts/v1.0.0.md
```diff
@@ -9,6 +9,7 @@ links:
 
 ## Changes affecting the Galasa Service
 
+- A new `Binary` credential type is now supported in the Galasa Credentials Store. Binary secrets store opaque binary data (for example, licence JAR files such as `my_license.jar`) that tests can retrieve at runtime via the Credentials Store. The `galasactl secrets set` command gains two new flags: `--binary-file` (reads a file from disk and base64-encodes its contents) and `--base64-binary-encoded` (accepts an already-encoded string). Binary secrets are mutually exclusive with all other credential fields. The REST API `PUT /secrets/{name}` and `GET /secrets/{name}` endpoints support the new `Binary` type alongside the existing types.
```
eamansour: I'm wondering whether `Binary` is the right name for this new type of secret - maybe it should be `Opaque` (like Kubernetes) or `File` if this secret is purely focused on storing opaque data that isn't necessarily in binary format?

Happy to keep as-is though!

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -3251,10 +3251,16 @@ components:
               description: |
                 The type of the KeyStore. Optional, defaults to 'PKCS12' if not specified.
                 Supported values are 'PKCS12' and 'JKS'. Only used when the Secret 'type' is set to 'KeyStore'.
+            data:
```
eamansour: Minor - would it be worth renaming this `data` field to `binaryData` or similar to avoid any awkwardness with the secret type's `data` field?

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
```diff
@@ -3399,6 +3405,20 @@ components:
           description: |
             The type of the KeyStore. Optional, defaults to 'PKCS12' if not specified.
             Supported values are 'PKCS12' and 'JKS'. Can be supplied on its own to update only the keystore type.
+        binary:
+          type: object
+          description: |
+            A binary value to create or update in the Credentials Store. Cannot be used alongside 'username',
+            'password', 'token', 'keystore', 'keystoreType' or 'keystorePassword'.
+          properties:
+            value:
```
eamansour, line 3414: Should this field be `data` (i.e. so it reads `binary.data`) to be more consistent with the `GalasaSecret` type?

modules/framework/galasa-parent/dev.galasa.framework.api.secrets/src/main/java/dev/galasa/framework/api/secrets/internal/routes/AbstractSecretsRoute.java
```diff
@@ -216,6 +225,34 @@ private ICredentials decodeTokenCredentials(SecretRequesttoken token) throws Int
         return new CredentialsToken(decodedToken);
     }
 
+    /**
+     * Decodes binary credentials from the request payload.
+     *
+     * The binary value must be base64-encoded. If encoding is "base64" the value
+     * is treated as double-encoded (base64-of-base64) and decoded once to yield
+     * the stored base64 representation.
```
eamansour: Slightly confused about this double-encoding assumption - if the secret is sent to the REST API in base64 encoding, doesn't that mean it's base64-encoded once?

Then we wouldn't need to decode binary credentials at all.

Perhaps we should also enforce that binary credentials need to have an encoding assigned?

modules/framework/galasa-parent/dev.galasa.framework.api.secrets/src/main/java/dev/galasa/framework/api/secrets/internal/routes/AbstractSecretsRoute.java
```diff
@@ -291,8 +331,13 @@ private void setSecretTypeValuesFromCredentials(GalasaSecretmetadata metadata, G
             data.setkeystore(encodeValue(keyStoreCredentials.getEncodedKeyStore()));
             data.setKeystorePassword(encodeValue(keyStoreCredentials.getKeyStorePassword()));
             data.setKeystoreType(keyStoreCredentials.getKeyStoreType());
-            
+
             metadata.settype(KEY_STORE);
+        } else if (secretType == GalasaSecretType.BINARY) {
+            ICredentialsBinary binaryCredentials = (ICredentialsBinary) credentials;
+            // getEncodedData() returns base64 without prefix, so we encode it again for the response
+            data.setdata(encodeValue(binaryCredentials.getEncodedData()));
```
eamansour: A bit confused here as well - if the binary data is given in base64 format, do we need to be encoding the data again?

eamansour (commented):
Just a few comments/questions, let me know if anything doesn't make sense!

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
eamansour: Other secret types use `value` so leaving this as `value`

jadecarino: As discussed on Slack, leaving this as `value` as otherwise it needs changing across all secret types and then the bean code would also need updating, making this change set much larger.

modules/cli/pkg/cmd/secretsSet.go
jadecarino: Yes great spot thanks, removed in https://github.com/galasa-dev/galasa/pull/670/commits/c6aa449f5604e7baca7110d78edf12765827656a

modules/cli/pkg/secrets/secretsSet.go
jadecarino: Thanks, removed in https://github.com/galasa-dev/galasa/pull/670/commits/c6aa449f5604e7baca7110d78edf12765827656a

modules/cli/pkg/cmd/secretsSet.go
jadecarino: Great question, the ETCD limit is 1.5MB, but I've found from testing `secrets set` on Minikube that anything over 760KB of raw data means a HTTP 413 is returned so I've added docs and an error message for data larger than 760 in https://github.com/galasa-dev/galasa/pull/670/commits/08e8f2fc8eecd509ca19d6e14a9528e5a7182f74

modules/framework/galasa-parent/dev.galasa.framework.api.secrets/src/main/java/dev/galasa/framework/api/secrets/internal/routes/AbstractSecretsRoute.java
jadecarino: Yes this is unnecessary on reflection so I have removed the double-encoding in https://github.com/galasa-dev/galasa/pull/670/commits/08e8f2fc8eecd509ca19d6e14a9528e5a7182f74

modules/framework/galasa-parent/dev.galasa.framework.api.openapi/src/main/resources/openapi.yaml
jadecarino: Yes good idea, as I've renamed the secret type name to Opaque, I have renamed this to `opaqueData` in https://github.com/galasa-dev/galasa/pull/670/commits/bd0a2afbc320d1201749145ab5b34267ceaba223 - I hope thats suitable?

docs/content/releases/posts/v1.0.0.md
jadecarino: Thanks for the suggestion, I've renamed to `Opaque` in https://github.com/galasa-dev/galasa/pull/670/commits/bd0a2afbc320d1201749145ab5b34267ceaba223

modules/framework/galasa-parent/dev.galasa.framework.api.secrets/src/main/java/dev/galasa/framework/api/secrets/internal/routes/AbstractSecretsRoute.java
jadecarino: Thanks yes I agree the double encoding assumption is wrong so I've removed it in https://github.com/galasa-dev/galasa/pull/670/commits/bd0a2afbc320d1201749145ab5b34267ceaba223 - however we don't want to enforce encoding on binary credentials (or set it at all) as otherwise it will attempt to decode it before storing it and we need it stored as a base64 representation

eamansour (approved):

PR #687, opened 2026-08-05
fix docker remson ssl cert + protocol issues

Mark-J-Lawrence:
## Why?

There were recent changes via https://github.com/galasa-dev/galasa/pull/551/ to introduce the ability to connect to TLS protected remote docker engines.

The support however didn't make it to the docker managers resource monitoring providers, so this PR delivers the ability to clean up docker resources on TLS protected servers.

It additionally delivers a fix, where if a http protocol was not set within the docker hostname CPS prop, an exception would be thrown in the resource monitoring providers when trying to construct the URI, due to it being invalid without the protocol.
The resource monitoring providers now follow the same logic as the Docker manager code itself and will automatically add a `http` or `https` depending on given factors.

## Changes
- [ ] Unit tests (if applicable)


jadecarino (approved):
Looks good, thanks for spotting this gap in my implementation good catch!

PR #703, opened 2026-08-24
Treat normal and secure built-in namespaces as existing namespaces

eamansour:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2631

This PR adds the `service` namespace as a built-in namespace.

Previously, all existing namespaces and non-existent namespaces would be initialised as `CPSNamespace` objects. The API server would then check if a namespace contained any properties to determine whether it actually exists. This meant that if a built-in namespace like `secure` or `service` didn't contain properties, then it would be treated as a non-existent namespace and return a 404 Not Found.

This PR fixes this by treating built-in namespaces as namespaces that always exist, regardless of whether they contain properties, so the API server will not return a 404 Not Found.

## Changes
- [x] Added `service` to the built-in namespaces


jadecarino (approved):
Apologies for the delay!

PR #708, opened 2026-09-11
fix(secrets): Add property separator to prefix deletions to stop other secrets from being deleted unintentionally

eamansour:
## Why?

Refer to https://github.com/galasa-dev/projectmanagement/issues/2642

## Changes
- [x] Added the property separator (".") to the deletePropertiesWithPrefix call so that properties with similar names do not get deleted unintentionally


jadecarino (approved):
Great spot!

